Issue #79, opened 2019-05-22
Schema issue with plist511_test


dragosprisaca:
The macos-definitions-schema.xsd schema includes the following mappings:

<oval:element_mapping>
    <oval:test>plist511_test</oval:test>
    <oval:object>plist511_object</oval:object>
    <oval:state>plist511_state</oval:state>
    <oval:item>plist_item</oval:item>
</oval:element_mapping>

This test should use a plist511_item.

Thanks,
_Dragos.

zport:
Issue resolved with PR #80 .

PR #80, opened 2019-05-22
Fix for https://github.com/OVAL-Community/OVAL/issues/79

solind:


solind:
Consensus call...

zport:
Hi David,

Looks like the consensus call was issued over 14 days ago... should we move this into develop?

Zack

solind:
Yes, but last I checked I do not have authority to do so with my account.  So, please merge and close.

zport:
Merged! Thanks @solind.

PR #81, opened 2019-05-30
File Attribute Type

wmunyan:
Added allowable empty string enumeration value for Entity(State|Item)FileAttributeType, per standard for other OVAL enumeration types.

### DIFF
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -9480,6 +9480,11 @@
                                 <xsd:documentation>This value is reserved for system use.</xsd:documentation>
                             </xsd:annotation>
                         </xsd:enumeration>
+                        <xsd:enumeration value="">
+                              <xsd:annotation>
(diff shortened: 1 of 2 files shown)


solind:
Good catch.  Was there an issue associated with this?

wmunyan:
@solind it was something we came across in our testing.  The OVAL results added an empty element with a status of "does not exist" and it failed schema validation.  Temporary workaround is to just not add the element to the `file_item`, since its a 0..unbounded, but wanted to keep consistency with other enums.

isimluk:
LGTM.

solind:
This is a good change because we need the state to be able to potentially accommodate a variable value.  However, I don't think it's actually possible for a Windows file to have no attributes, as in that case it should have FILE_ATTRIBUTE_NORMAL, which is defined as:

"A file that does not have other attributes set. This attribute is valid only when used alone."

solind:
This also needs to be closed and changed into a pull request to merge into the develop branch.

PR #83, opened 2019-05-30
Proposal:  Addition of the pwpolicy (5.12) test to Mac OSX schema

wmunyan:
Added new OSX pwpolicy-based test to account for new command format (using -getaccountpolicies and XML output) and deprecated options (-getpolicy)

### DIFF
--- a/oval-schemas/macos-definitions-schema.xsd
+++ b/oval-schemas/macos-definitions-schema.xsd
@@ -2193,7 +2193,7 @@
       <!-- =============================================================================== -->
       <xsd:element name="pwpolicy512_test" substitutionGroup="oval-def:test">
             <xsd:annotation>
-                  <xsd:documentation>This test retrieves password policy data from the 'pwpolicy -getaccountpolicies -u target_user [-a username] [-p userpass] [-n directory_node]' output where username, userpass, and directory_node are optional. Please see the 'pwpolicy' man page for additional information.</xsd:documentation>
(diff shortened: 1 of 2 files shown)


solind:
@wmunyan do you know which version of macOS/Darwin added the getaccountpolicies pwpolicy command?

Also... do you think there should also be an object entity for specifying the "authenticator" name (the -a command-line option)?

wmunyan:
I _believe_ the 10.13 (High Sierra?) version introduced the new option and deprecated the old ones.  Admittedly I am having issues finding specific 10.13 man pages for it, but have found it in 10.14.

For the authenticator element, per previous versions of the test, the `username` element is the value to use with the `-a` option.  I am open to changing that element name to `authenticator`, if that is clearer.

solind:
Actually... never mind, I guess the "username" entity is supposed to be the authenticator name.

What would you think about re-naming the entities from target_user->username, and username->authenticator, to more reasonably align with the command-line arguments for the pwpolicy command?  I know the legacy pwpolicy test objects/states/items used the target_user/username convention, but I always thought it was a bit confusing (as evidenced by my question!)

wmunyan:
agreed.  i have no problems changing those element names.  I will update!

wmunyan:
OK - Updated.  Feedback welcome

solind:
Hey, sorry for failing to comment on this way back.  This looks fine to me.  This is a consensus call to start the 2-week count-down.

solind:
And this also needs to be closed and changed into a merge request into the develop branch.

PR #85, opened 2019-06-13
Update Consensus Building Information

zport:
There was no instruction given to Area Supervisors previously on how to initiate a Consensus Call. This pull request documents the process used by two Area Supervisors up to this point, which involves sending a descriptive email out to the OVAL mailing lists asking for final feedback.


adammontville:
Generally, this is a good change. I do have two comments:
1. The prose instructs Area Supervisors to send an email to the "OVAL mailing lists" and provides a link to the page describing all three of our lists, one of which is closed (OVAL Board). Should we name the lists to which to send the email? (Should they always go to OVAL Developer and OVAL Repository?)
2. This may be pedantic, but do we want to specify a specific time and timezone 14 days from the date the consensus call is initiated? If we do not, then a late breaking objection could come in after the PR is processed and still be, for the objector, within the 14-day period. (This may be a bit of an edge case, but I would not be opposed to saying something like 11:59 PM UTC on date X.)

DavidRies:
Hi Zack, I second--this looks good. I'd say the email should go to the OVAL Developer List only (repo is for content/repo issues).

zport:
Updated with the changes recommended above, as well as a couple of spelling errors I caught. Any other thoughts? Thanks!

PR #88, opened 2019-06-13
Updated proposal; Add 2 new Linux Tests

wmunyan:
Modified original proposal.  Removed `lsmod` and `modprobe`, replaced with single `kernelmodule` construct.  

The 2nd test in the PR is a new construct, proposing the `sestatus_(test|object|state|item)`.  This test allows implementations to collect and evaluate information stemming from the output of the `sestatus` command.

### DIFF
--- a/oval-schemas/linux-definitions-schema.xsd
+++ b/oval-schemas/linux-definitions-schema.xsd
@@ -2376,12 +2376,17 @@
                                         <xsd:documentation>This indicates whether SELinux is currently enforcing the policies or not utilizing the following values enforcing, permissive, disabled.</xsd:documentation>
                                    </xsd:annotation>
                               </xsd:element>
-                              <xsd:element name="mode_from_config" type="linux-def:EntityStateSEStatusModeType" minOccurs="0" maxOccurs="1">
+                              <xsd:element name="mode_from_config_file" type="linux-def:EntityStateSEStatusModeType" minOccurs="0" maxOccurs="1">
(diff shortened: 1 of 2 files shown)


solind:
Hi @wmunyan, why make selinux_status an EntityItemStringType instead of an EntityItemSEStatusModeType?  The only possible values should be enabled and disabled, right?

wmunyan:
@solind I think what I'll do is add another enumeration for `EntityItemSELinuxStatusType` with values for `enabled` and `disabled`.  The `Mode` type includes the `enforcing` and `permissive`, etc which don't match up for the status.

isimluk:
Looks good to me.

solind:
If this can be changed into a PR for the develop branch, @isimluk can make a consensus call so we can move forward and merge.

isimluk:
Consensus call initiated: http://lists.cisecurity.org/pipermail/oval_developer_lists.cisecurity.org/Week-of-Mon-20200127/000615.html

isimluk:
Consensus call period elapsed. None of the feedback received was identified to be required to inform changes to this PR.

This is good to merge. However, we may want to consider merging this first to the `develop` branch as @solind suggested.

solind:
Hi @isimluk, it's not a suggestion.  After the consensus call, the community process calls for merging into the development branch.

https://oval-community-guidelines.readthedocs.io/en/latest/proposal-process/release-process.html#release-process

PR #113, opened 2021-02-05
Introduce a schema directory README that points out to oval-community-guidelines.readthedocs.io

matejak:
Having more than one "official" place that contains schemas looks like asking for trouble - when schemas are updated, there is a risk that one of those repositories will fall behind and mislead others.

I have also come across https://github.com/OVALProject/Language/tree/master/docs that contains human-readable OVAL schema  documentation in form of Markdown files that is extracted from XSDs. What is the relation of the OVALProject to CIS/community OVAL repositories?

### DIFF
--- a/oval-schemas/README.md
+++ b/oval-schemas/README.md
@@ -1,0 +1,2 @@
+This directory contains authoritative OVAL schemas in form of XSD files.
+If you are after a human-readable description of those schemas, check out [OVAL Schemas on readthedocs.io](https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/index.html) that are available in the HTML form.


solind:
Hi Matej,

The current “official” home of the OVAL schemas is: https://github.com/OVAL-Community/OVAL <https://github.com/OVAL-Community/OVAL>

And, the “official” home-page for OVAL is: https://oval-community-guidelines.readthedocs.io/en/latest/index.html <https://oval-community-guidelines.readthedocs.io/en/latest/index.html>

The readthedocs site also has schema documentation (although in my opinion it’s not as pretty or usable as the original OVALProject markdown schema documentation).  Whoever maintains that OVALProject Git repo should probably point to the newer home for the language.

Best regards,
—David Solin

matejak:
OK, so removing those schemas probably wasn't a good idea. So I propose to add a README file that point to the generated documentation on the rtd website.
One question remains though - if official schemas are those in this repository, what kind of schemas are hosted here: https://github.com/CISecurity/OVALRepo/tree/master/oval_schemas

Issue #118, opened 2021-03-17
Proposal: Add new OVAL schema for AWS


wmunyan:
**Abstract**
Recently released security best practice guidance for cloud service providers has necessitated the creation of OVAL constructs which interface with those CSPs.  Because the CSPs have constructed and released API's for the management of tenant resources, this proposal intends to "OVALize" the API's read-only methods in order to enable the collection and evaluation of cloud-native security policies.

**Link to Proposal**
https://github.com/OVAL-Community/OVAL/pull/123

**Additional context**
We currently want to solicit assistance in the development of a proposed schema/format of the `xml_output` field in the `api_item` system characteristics construct.

Currently this element is specified as an "any XML" type, and building consensus on a standard schema/format for those results would help interoperability.  However, standard output from the AWS CLI is JSON, and the SDK is built with object representations of output.  If a generic XML representation can be developed by the community that would be excellent.

If the system characteristics are to be represented as `record` datatypes containing multiple `field` elements, a proposal would need to be included that allows for `field` elements to allow their own `record` datatypes, thus nesting `record`s within `field`s and so forth (see the very old issue #13).

Discussion on this point is most welcome.  If sample content (and perhaps some proposed output) would help visualize, we're happy to try and propose some.

solind:
Hey @wmunyan, why not let the output field contain the JSON output, and add a "JSONPath" type of entity to the object(s) to permit direct querying?

There is also the possibility of somehow making variable values JSONPath-queryable.

wmunyan:
@solind I have absolutely no issue if the output field contains JSON.  How would we go about specifying that as a datatype and shimmying JSON data into an XML element?  I searched around looking for examples, but it could just be the JSON formatted as text and using an EntityItemStringType?

If we do go this route, I would like to propose that any JSONPath being used conforms to the specification being proposed in the IETF JSONPath working group -- https://datatracker.ietf.org/doc/draft-ietf-jsonpath-base/.  The working group stemmed from an already existing set of rules/implementations, and is being formalized in the IETF now.  As the specification is still a draft, it is still subject to change, but I'd just like it to adhere to some sort of standard if possible.

wmunyan:
I've updated the PR and the schemas to incorporate the output field being JSON and added a JSONPath element to query the output for the required information.

I also added a second test for `apicontent` which is similar to textfilecontent (regex searching) for API calls that do not produce JSON output, such as `aws s3 ls`.

[...]

PR #120, opened 2021-03-17
Proposal: Add schema for AWS

wmunyan:

As per the related issue (#118) the specification of the output_xml field is up for discussion as well; whether it should be up to tool vendors to implement some sort of API output-to-XML translation, or if further proposals are to be added expanding the record datatype. See the issue for more information and to add to the discussion.


wmunyan:
i have no idea why this is thinking all these changes need to be merged as part of this PR....

wmunyan:
Closing this because it isnt comparing the right things to the right things.

PR #121, opened 2021-03-24
Adding AWS-API schemas for new proposal (Take 3)

wmunyan:

This PR includes 3 test/obj/state constructs:

- `api_test`:  The API test allows for the invocation of the AWS API for those operations that return JSON content, and allows collection of specific data
- `apicontent_test`:  The API CONTENT test allows for the invocation of the AWS API for those operations that do not return JSON content, but plain text, such as the `aws s3 ls` operation.  This object contains a regular expression pattern similar to that of the textfilecontent object, to parse the output.
- `credentialreport_test`:  The Credential Report test is a very specific test with regards to generating and parsing the output of the AWS credential report.  More info in the schema documentation on that.

This updates the API test according to @solind's comment regarding using an element for holding JSON data and using JSONPath expressions to evaluate collection of information.


DavidRies:
Hi @wmunyan , 

I reviewed the new tests and... generally this all looks great! I'm excited to have these capabilities in OVAL. 

Here are a few minor comments...

### Versioning?

Because the `api_test` and `apicontent_test` tests require content authors to encode specific AWS API methods in the content as well as expectations of specific response data structures, content may break (or yield unintended results) when the API changes. 

Should we add an API version (required for all objects, optional for state, included in item)? That would allow the engine to attempt to use the API version the content author wrote against. If/when the API version is no longer available, that would also make it much easier to identify the affected objects. WDYT?

### API Orientation of `api_test` and `apicontent_test`

Generally speaking, I prefer OVAL tests that are configuration/control-oriented over those that are API/protocol-oriented like these.  When crafted properly, they can make the content easier to write (and read) and can allow the engines to absorb the complexities of the APIs and result data structures. 

For example, it might be nicer to have a `instance_security_group_test` that allowed authors to make assertions about the protocols allowed for ingress/egress to instances. With the protocol-oriented tests proposed, I think this is still possible, but it might require making calls to one API method to get instances and associated security group ids which are used as a filter on the results of another object/API call to get security group ingress/egress properties.

However, I do think this protocol-orientation is the right approach right now. The API is so extensive that creating simplified, semantic OVAL tests for common use cases would take a long time. And, because the API and security controls are evolving relatively quickly, we'd need this generic/API capability in any case.

I hope that over time, we--as a community--will discover common usage patterns of these API calls and will add simplified, semantic tests for them.

### `credentialreport_test`

I'm not familiar with this report or the types of assertions that would typically made against it. Having `access_key_1_*` entities along with `access_key_2_*` entities might be inconvenient. 

For example, if I wanted to make an assertion about the last rotated date, I would need to make it twice, once for `access_key_1_last_rotated` and once for `access_key_2_last_rotated`. The same goes for the `cert_1` and `cert_2` entities.

Do you think it would make sense to break this into three tests as follows:

1) credentialreportsuser_test with state/item entities: user, arn, user_creation_time, password_enabled, password_last_used, password_last_changed, password_next_rotation, mfa_active
2) credentialreportskey_test with state/item entities: user, arn, access_key_active, access_key_last_rotated, access_key_last_used_date, access_key_last_used_region, access_key_last_used_service
3) credentialreportscert_test with state/item entities: user, arn, cert_active, cert_last_rotated

That way the credentialreportskey_object would return one or two items depending on whether there were one or two keys for the user.

Thoughts?

-David

PR #124, opened 2021-05-26
Proposal to add schema for Palo Alto

wmunyan:
The single proposed test (config_test) is used to check the properties of the XML output from a PAN-OS XML API request to export the current running configuration. This is a request to the API at `https://[PAN-OS-DEVICE]/api/?type=export&amp;category=configuration`.  

The response to this request is an XML payload rooted with a "response" element and including device-specific information.


solind:
First comment... should this really be the paloalto schema? I think panos schema would be more consistent with the naming convention for other router OSes.

PR #163, opened 2024-09-30
Updated terms of use based on current guidance

vanderpol:
Replacing the old terms of use, which mentioned CIS with a generic government license, recommended by NIWC IP lawyers


harrisdk (approved):
License/term-of-use updated based on guidance from NIWC Legal

PR #171, opened 2024-10-03
156 add trustee names back to windows fileeffectiverights item user sid item and group sid item

vanderpol:
Un-deprecated trustee name from windows file/registryeffectiverights and file/registryauditing items
Added user/group/subgroup back to windows user_sid_item and group_sid_item


vanderpol:
I think I made a mistake with this branch/commit, going to close and try again.

PR #177, opened 2024-10-10
Add Merge function type and sort enumeration

vanderpol:
Adding a merge function which joins together multiple elements into a single variable, with an optional delimiter, and optional sorting.  Had to add a new sort enumeration which may need to be discussed.

### DIFF
--- a/oval-schemas/oval-definitions-schema.xsd
+++ b/oval-schemas/oval-definitions-schema.xsd
@@ -941,13 +941,14 @@
     </xsd:complexType>
     <xsd:complexType name="MergeFunctionType">
         <xsd:annotation>
-            <xsd:documentation>The merge function takes one or more components and merges them together into a single string, with an optional delimiter.  For example if data from one registry multi-size value contains values of "abc" and "def".  The merge function will resolve to a local_variable with the value of 'abcdef'.  If an optional delimiter of ',' was used, the merge function would resolve to a local_variable with the value of 'abc,'def'</xsd:documentation>
+            <xsd:documentation>The merge function takes one or more components and merges them together into a single string value, optionally including a delimiter string. For example, if data from one registry multi-size value contains values of "abc" and "def", the merge function operated on those values would resolve to a string with the value of 'abcdef'. If an optional delimiter of ',' was used, the merge function would resolve to a value of 'abc,'def'.</xsd:documentation>
         </xsd:annotation>
         <xsd:sequence minOccurs="1" maxOccurs="unbounded">
             <xsd:group ref="oval-def:ComponentGroup"/>
         </xsd:sequence>
 		<xsd:attribute name="delimiter" type="xsd:string" use="optional" default=""/>
-		<xsd:attribute name="sort" type="oval-def:SortEnumeration" use="optional" default="none"/>
+		<xsd:attribute name="sort" type="oval-def:SortEnumeration" use="optional" default="document"/>
+		<xsd:attribute name="order" type="oval-def:OrderEnumeration" use="optional" default="ascending"/>
     </xsd:complexType>
     <xsd:complexType name="CountFunctionType">
         <xsd:annotation>
@@ -1116,27 +1117,44 @@
     </xsd:simpleType>
     <xsd:simpleType name="SortEnumeration">
         <xsd:annotation>
-            <xsd:documentation>The SortEnumeration simple type defines basic sorting operations.  Currently 'none', 'alphabetical' and 'alphanumerical' are defined..</xsd:documentation>
+            <xsd:documentation>The SortEnumeration simple type defines basic sorting operations.  Currently 'document', 'lexical', 'numeric' and 'natural' are defined.</xsd:documentation>
         </xsd:annotation>
         <xsd:restriction base="xsd:string">
-            <xsd:enumeration value="none">
+            <xsd:enumeration value="document">
                 <xsd:annotation>
-					<xsd:documentation>No sorting is performed, the values will remain in the order they came in as.</xsd:documentation>
+					<xsd:documentation>No sorting is performed, the values will remain in the order they came in as, and match the order in which the referenced system characterics will appear in the OVAL results XML document.</xsd:documentation>
                 </xsd:annotation>
 			</xsd:enumeration>	
-            <xsd:enumeration value="alphabetical">
+            <xsd:enumeration value="lexical">
 			    <xsd:annotation>
-                    <xsd:documentation>Sort alphabetical, useful for pure string lists</xsd:documentation>
+                    <xsd:documentation>Sort alphabetically, useful for pure string lists.</xsd:documentation>
                 </xsd:annotation>
 			</xsd:enumeration>
-			<xsd:enumeration value="alphanumerical">
+			<xsd:enumeration value="numeric">
 			    <xsd:annotation>
-                    <xsd:documentation>Sort alphanumerical, useful for lists of numbers or lists of strings with numbers</xsd:documentation>
+                    <xsd:documentation>Sort numerically, useful for lists of numbers or lists of integers or floats, but will cause errors if string data is present.</xsd:documentation>
                 </xsd:annotation>
 			</xsd:enumeration>
 			<xsd:enumeration value="natural">
 			    <xsd:annotation>
                     <xsd:documentation>Sort like PHP's natsort https://www.php.net/manual/en/function.natsort.php</xsd:documentation>
+                </xsd:annotation>
+			</xsd:enumeration>
+        </xsd:restriction>
+    </xsd:simpleType>
+    <xsd:simpleType name="OrderEnumeration">
+        <xsd:annotation>
+            <xsd:documentation>The OrderEnumeration simple type defines the direction in which data should be sorted, and works to support the SortEnumeration.  The OrderEnumeration has no impact on the SortEnumeration of 'document'.</xsd:documentation>
+        </xsd:annotation>
+        <xsd:restriction base="xsd:string">
+            <xsd:enumeration value="ascending">
+                <xsd:annotation>
+					<xsd:documentation>Numbers: Arranged from smallest to largest, such as 0, 1, 2, 3, 4, 5, 6. Letters: Arranged alphabetically from A to Z</xsd:documentation>
+                </xsd:annotation>
+			</xsd:enumeration>	
+            <xsd:enumeration value="descending">
+			    <xsd:annotation>
+                    <xsd:documentation>Numbers: Arranged from largest to smallest, such as 9, 8, 7, 6. Letters: Arranged alphabetically from Z to A</xsd:documentation>
                 </xsd:annotation>
 			</xsd:enumeration>
         </xsd:restriction>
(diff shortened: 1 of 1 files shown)


oval-schemas/oval-definitions-schema.xsd
```diff
@@ -938,6 +939,16 @@
             <xsd:group ref="oval-def:ComponentGroup"/>
         </xsd:sequence>
     </xsd:complexType>
+    <xsd:complexType name="MergeFunctionType">
+        <xsd:annotation>
+            <xsd:documentation>The merge function takes one or more components and merges them together into a single string, with an optional delimiter.  For example if data from one registry multi-size value contains values of "abc" and "def".  The merge function will resolve to a local_variable with the value of 'abcdef'.  If an optional delimiter of ',' was used, the merge function would resolve to a local_variable with the value of 'abc,'def'</xsd:documentation>
```
solind, line 944: Edited slightly:

The merge function takes one or more components and merges them together into a single string value, optionally including a delimiter string. For example, if data from one registry multi-size value contains values of "abc" and "def", the merge function operated on those values would resolve to a string with the value of 'abcdef'. If an optional delimiter of ',' was used, the merge function would resolve to a value of 'abc,'def'.

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -1103,6 +1114,33 @@
             -->
         </xsd:restriction>
     </xsd:simpleType>
+    <xsd:simpleType name="SortEnumeration">
+        <xsd:annotation>
+            <xsd:documentation>The SortEnumeration simple type defines basic sorting operations.  Currently 'none', 'alphabetical' and 'alphanumerical' are defined..</xsd:documentation>
+        </xsd:annotation>
+        <xsd:restriction base="xsd:string">
+            <xsd:enumeration value="none">
```
solind, line 1122: I wouldn't call this "none", I would call this "document_order".  After all, two interpreters using the same system-characteristics file as input should both generate identical results.

vanderpol: I wasn't entirely sure what your "document_order" meant, as long as we explain it as the order of the items/values in  OVAL system characteristics, I would agree.

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -1103,6 +1114,33 @@
...
+                <xsd:annotation>
+					<xsd:documentation>No sorting is performed, the values will remain in the order they came in as.</xsd:documentation>
+                </xsd:annotation>
+			</xsd:enumeration>	
+            <xsd:enumeration value="alphabetical">
```
solind, line 1127: I'd call this "lexical", which I believe is more standard terminology, and typically stands in contrast with "natural" ordering.

For example, given the strings {"a1, "a2", "a12", "1", "2", "12"}, in ascending order*, lexical sorting yields:

"1", "12", "2", "a1", "a2", "a12"

And natural sorting yields:

"1", "2", "12", "a1", "a12", "a2"

* this reminds me, there is a difference between the sort type and the order (ascending vs. descending)

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -1103,6 +1114,33 @@
...
+			    <xsd:annotation>
+                    <xsd:documentation>Sort alphabetical, useful for pure string lists</xsd:documentation>
+                </xsd:annotation>
+			</xsd:enumeration>
+			<xsd:enumeration value="alphanumerical">
```
solind, line 1132: How is this different from "natural"?

oval-schemas/oval-definitions-schema.xsd
vanderpol: I'm fine with that, will update

solind (changes requested):
I think maybe you need a SortEnumeration (lexical, natural, numeric) and an OrderEnumeration (ascending/descending)

Where numeric would generate an error if it encounters a string value that cannot be converted to a base-10 number (float).

oval-schemas/oval-definitions-schema.xsd
solind: Document order means... this is an XML document.  The order is the order in which the associated elements appear in the document.

vanderpol:
> I think maybe you need a SortEnumeration (lexical, natural, numeric) and an OrderEnumeration (ascending/descending)
> 
> Where numeric would generate an error if it encounters a string value that cannot be converted to a base-10 number (float).

I wasn't sure about a numeric sort, as accidental string data could cause all kinds of errors, but an alphanumeric (or natural), would handle them both.  I'm not 100% certain on the need for alphanumeric vs natural, likely just natural.   I can easily add, and content authors beware.

Agree on OrderEnumeration, will add.

solind:
As for numerical vs. natural, it's just a thought.  You could effectively achieve the same thing using a variable datatype, like you did. 😀

oval-schemas/oval-definitions-schema.xsd
vanderpol: In my brain, this all happens in memory structures, long before we end up writing it to XML, so that was my disconnect.

vanderpol:
So for the OrderEnumeration, I'm pondering the interaction between that and "document" sort?  I guess we can just document that the OrderEnumeration has no influence on the 'document' sort, as 'document' sort, isn't sorting at all?  Default OrderEnumeration should be set to "ascending".

vanderpol:
Here's some updated sample results, I've added a recursive Registry test, a numeric sort test using variables, and I threw in a WMI57 test to show it works with records.
[SCC-5.10.1_Beta1_2024-10-11_084515_OVAL-Results_merge_test_content.xml.txt](https://github.com/user-attachments/files/17342769/SCC-5.10.1_Beta1_2024-10-11_084515_OVAL-Results_merge_test_content.xml.txt)

solind:
> So for the OrderEnumeration, I'm pondering the interaction between that and "document" sort? I guess we can just document that the OrderEnumeration has no influence on the 'document' sort, as 'document' sort, isn't sorting at all? Default OrderEnumeration should be set to "ascending".

OrderEnumeration values would be "ascending" and "descending".  I guess in the case of a "document_order" sort value, this would be ignored.

solind:
> Here's some updated sample results, I've added a recursive Registry test, a numeric sort test using variables, and I threw in a WMI57 test to show it works with records. [SCC-5.10.1_Beta1_2024-10-11_084515_OVAL-Results_merge_test_content.xml.txt](https://github.com/user-attachments/files/17342769/SCC-5.10.1_Beta1_2024-10-11_084515_OVAL-Results_merge_test_content.xml.txt)

Naturally, a little digression here... I noticed this item in your results:

```
<win-sc:registry_item id="10" status="exists">
  <win-sc:hive>HKEY_LOCAL_MACHINE</win-sc:hive>
  <win-sc:key>SYSTEM\CurrentControlSet\Control\Session Manager</win-sc:key>
  <win-sc:name>ExcludeFromKnownDlls</win-sc:name>
  <win-sc:last_write_time datatype="int">133730601810000000</win-sc:last_write_time>
  <win-sc:type>reg_multi_sz</win-sc:type>
  <win-sc:value></win-sc:value>
  <win-sc:value></win-sc:value>
  <win-sc:windows_view>64_bit</win-sc:windows_view>
</win-sc:registry_item>
```

I see two empty values, which in XML parlance means strings of zero length.  I assume this happened because a REG_SZ is terminated by a pair of NULL (0x00) bytes.  IIRC, Joval would interpret this as a non-valued key, and represent that by creating a single item value like this:

```
  <win-sc:value status="does not exist"/>
```

This is important, because empty strings are still string values that exist.

vanderpol:
Very astute observation in the results, and given the lack of multi-sz content in the wild we may have overlooked this.  Below is an export from my registry, so just to confirm you would say that this should be a DNE on the value for ExcludeFromKnownDlls?

Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager]
"AutoChkTimeout"=dword:00000008
"BootExecute"=hex(7):61,00,75,00,74,00,6f,00,63,00,68,00,65,00,63,00,6b,00,20,\
  00,61,00,75,00,74,00,6f,00,63,00,68,00,6b,00,20,00,2a,00,00,00,00,00
"BootShell"=hex(2):25,00,53,00,79,00,73,00,74,00,65,00,6d,00,52,00,6f,00,6f,00,\
  74,00,25,00,5c,00,73,00,79,00,73,00,74,00,65,00,6d,00,33,00,32,00,5c,00,62,\
  00,6f,00,6f,00,74,00,69,00,6d,00,2e,00,65,00,78,00,65,00,00,00
"CriticalSectionTimeout"=dword:00278d00
"ExcludeFromKnownDlls"=hex(7):00,00

vanderpol:
@solind any other thoughts on the PR?  I feel like (after a lot of improvements based on your feedback, which I greatly appreciate) that we are pretty much good to go on this?

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -938,6 +939,17 @@
             <xsd:group ref="oval-def:ComponentGroup"/>
         </xsd:sequence>
     </xsd:complexType>
+    <xsd:complexType name="MergeFunctionType">
+        <xsd:annotation>
+            <xsd:documentation>The merge function takes one or more components and merges them together into a single string value, optionally including a delimiter string. For example, if data from one registry multi-size value contains values of "abc" and "def", the merge function operated on those values would resolve to a string with the value of 'abcdef'. If an optional delimiter of ',' was used, the merge function would resolve to a value of 'abc,'def'.</xsd:documentation>
```
solind, line 944: REG_MULTI_SZ doesn't actually mean "multi-size", it means multiple null-terminated strings.

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -1103,6 +1115,50 @@
...
+                </xsd:annotation>
+			</xsd:enumeration>
+			<xsd:enumeration value="numeric">
+			    <xsd:annotation>
+                    <xsd:documentation>Sort numerically, useful for lists of numbers or lists of integers or floats, but will cause errors if string data is present.</xsd:documentation>
+                </xsd:annotation>
+			</xsd:enumeration>
+			<xsd:enumeration value="natural">
+			    <xsd:annotation>
+                    <xsd:documentation>Sort like PHP's natsort https://www.php.net/manual/en/function.natsort.php</xsd:documentation>
```
solind, line 1140: Natural sort order is a pretty established thing, with implementations in most programming languages and on most operating systems.  So, I think this might be a better external URL to reference: https://en.wikipedia.org/wiki/Natural_sort_order

solind (commented):
My only comments are documentation-related.  LMK if you agree with them.

oval-schemas/oval-definitions-schema.xsd
vanderpol: I swear I know what I'm doing from time to time... probably would have been faster to have you write this spec from the start.

oval-schemas/oval-definitions-schema.xsd
vanderpol: Sounds good.

vanderpol:
Fixed documentation per your suggestions.  Hoping this is now done.

solind (approved):
Looks good to me!

vanderpol:
Just realizing I never attached any 'clean' test content to this PR.  Please see attached 
[merge_test_content.xml.txt](https://github.com/user-attachments/files/18510311/merge_test_content.xml.txt)

Issue #153, opened 2024-09-20
Add independent SQL512 test, deprecating SQL57


vanderpol:
**Abstract**
NIWC has developed SQL Server content using the SQLEXT OVAL specification previously submitted to OVAL https://lists.cisecurity.org/empathy/thread/KP4JJ3BINRY3VSDH6FP7UQWYDM2G4IW5

We realize that the SQLEXT OVAL test is non-standard, as it provides the content author the ability to have the application create sets of results (xccdf/OVAL etc..) per host, per DBMS instance or per database, inside the OVAL object.

As the SQLEXT is used by SCC and has publicly available content, we would like to find a way to make this test officially part of SCAP.  https://www.niwcatlantic.navy.mil/Technology/SCAP/scap-content-repository/

Our proposal is to move the decision point on when/if to create results based on 'host' or 'instance' or 'database to XCCDF 1.3 as part of SCAP 3.0.
- SCAP 3.0 Benchmark could have a new 'target' element in the XCCDF document which could be "host", "dbmsinstance", "database", "webserver", "website", default would be "host" and creating results based on the specified target would be optional.
- the SQLEXT (renamed as SQL512), could then have it's "behavior" element removed, making the OVAL object be 'normal'

**Link to Proposal**
Once a proposal has been put into pull request form, add a link to the PR here. If and as alternate proposals or objections are added they should be linked here as well.

**Additional context**
Add any other context or screenshots about the enhancement.

vanderpol:
Potential XCCDF 1.3 spec updates to allow for benchmarks to define the scope for scanning, and result creation.  


`````
<xsd:element name="target" type="cdf:targetType" minOccurs="0" maxOccurs="1" default="host">
     <xsd:annotation>
          <xsd:documentation xml:lang="en">For content targeting technology other than the entire operating system, content authors can specify the target type.  This selection may cause the SCAP processing application to create more than one set of results, one per target instance.</xsd:documentation>
     </xsd:annotation>
</xsd:element>



<xsd:simpleType name="targetType">
	<xsd:annotation>
		<xsd:documentation xml:lang="en"> The targetType represents the technology of the target that is in scope for being scanned, and can be a basis for creating separate instances of XCCDF results per target type.
		</xsd:documentation>
	</xsd:annotation>
	<xsd:restriction base="xsd:string">
		<xsd:enumeration value="host">
			<xsd:annotation>
[...]

PR #179, opened 2024-10-16
Added sql512 and deprecated sql57 based on #153

vanderpol:
Attached are sample results from the SCC application, which has implemented this test.  
[SQL512_Sample_Results.zip](https://github.com/user-attachments/files/17396752/SQL512_Sample_Results.zip)

These results are from a SCAP datastream, and contain more than SQL512 tests, but there are plenty of SQL512 tests to demonstrate it's functionality.  

This method is generating results for the entire host.  IF/when SCAP 3.0 implements a 'target' method, results could then be generated based on host/instance/database.

### DIFF
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -1408,12 +1408,16 @@
                                                 </xsd:element>
                                                 <xsd:element name="instance" type="oval-def:EntityObjectStringType">
                                                       <xsd:annotation>
-                                                            <xsd:documentation>The instance entity defines the specific instance name to be used when connecting to the correct database.</xsd:documentation>
-                                                      </xsd:annotation>
+                                                            <xsd:documentation>The instance entity defines the specific instance name to be used when connecting to the correct database, where instance refers to the running instance of the DMBS software itself. This could be a separate installation of binaries (such as with MS SQL Server), or just a set of running processes used to manage the DBMS.</xsd:documentation>
+															<xsd:documentation>The OVAL interpreter will automatically determine the list of available instances on the target.</xsd:documentation>
+															<xsd:documentation>When a pattern or string is entered, the OVAL interpeter will consider any matching instance as in scope for analysis.</xsd:documentation>
+													  </xsd:annotation>
                                                 </xsd:element>
                                                 <xsd:element name="database" type="oval-def:EntityObjectStringType" nillable="true">
                                                       <xsd:annotation>
-                                                            <xsd:documentation>The database entity defines the specific database name to be used when connecting to the specified instance. If the xsi:nil attribute is set to true, then the target instance's default database will be queried.</xsd:documentation>
+                                                            <xsd:documentation>The database entity defines the specific database name to be used when connecting to the specified instance, where a database is defined as a collection of tables within a DBMS instance.</xsd:documentation>
+															<xsd:documentation>When a pattern or string is entered, the OVAL interpeter will perform the query against any matching databases.</xsd:documentation>
+															<xsd:documentation>If the xsi:nil attribute is set to true, then only the target instance's default database will be queried.</xsd:documentation>
                                                       </xsd:annotation>
                                                 </xsd:element>
                                                 <xsd:element name="sql" type="oval-def:EntityObjectStringType" nillable="true">
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -1327,8 +1327,32 @@
       <!-- ================================================================================ -->
       <xsd:element name="sql512_test" substitutionGroup="oval-def:test">
             <xsd:annotation>
-                  <xsd:documentation>The sql512 test is used to check information stored in a database. It extends the standard TestType as defined in the oval-definitions-schema and one should refer to the TestType description for more information.  This test should only be performed by the OVAL scanner if the content is 'trusted', such as being digitally signed by a trusted content author.</xsd:documentation>
-                  <xsd:appinfo>
+                  <xsd:documentation>The sql512 test is used to check information stored in a database. It extends the standard TestType as defined in the oval-definitions-schema and one should refer to the TestType description for more information.</xsd:documentation>
+				  <xsd:documentation>This test should only be performed by the OVAL interpreter if the content is 'trusted', such as being digitally signed by a trusted content author.</xsd:documentation>
+                  <xsd:documentation>The OVAL interpeter will provide all authentication capabilities to the SQL DMBS target.</xsd:documentation>
+				  <xsd:documentation>The OVAL interpeter will query the target system and find all applicable DBMS instances and databases (refer to sql512 object elements for more information on instances and databases) .</xsd:documentation>
+				  <xsd:documentation>Using Microsoft SQL Server as an example, below is sample of what the OVAL intepreter will gather from a target.</xsd:documentation>
+				  <xsd:documentation>Target Host: Host1
+    SQL Server Instances:
+		SQLEXPRESS (version 13.0.6450.1 )
+			Databases:
+				master
+				model
+				msdb
+				tempdb
+				userdb1
+				userdb2
+		SQLSERVER (version 16.0.4135.4)
+			Databases:
+				master
+				model
+				msdb
+				tempdb
+				testdb1
+				testdb2		
+</xsd:documentation>
+<xsd:documentation>Content can then be created that targets one or more versions, and within those versions, queries could be run against one or more instances and one ore more databases.</xsd:documentation>
+				  <xsd:appinfo>
                         <oval:element_mapping>
                               <oval:test>sql512_test</oval:test>
                               <oval:object>sql512_object</oval:object>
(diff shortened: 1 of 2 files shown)


solind:
I don't think it's clear what the distinction is between the "instance" and "database" entities, as the documentation presupposes meanings for these terms (their meanings are not immediately clear to me either), and in your example the items all share the same instance "SQLEXPRESS" and the database value of "" (empty string).

Do you want to add meanings or examples to the documentation?  My guess is that an "instance" means a distinct running instance of the DBMS, a server process listening on a distinct TCP port (i.e., a "SID" on Oracle).  Whereas a "database" denotes something like a "tablespace" on Oracle.

vanderpol:
Good point @solind, I was just re-using the schema files from a previous developer, and should have put more into these new data elements.  I'm no SQL Server/Oracle guru either, but from my view the following are true:

- instance = The DBMS management system itself and any databases that it may contain.  This could be a separate installation of binaries, or just a set of running processes used to manage the DBMS.
- database = A specific database name within the DMBS system, where a database is a collection of tables

MS documentation on databases:  https://learn.microsoft.com/en-us/sql/relational-databases/databases/databases?view=sql-server-ver16

MS documentation on database engine instances:
https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/database-engine-instances-sql-server?view=sql-server-ver16

Oracle Instance docs:
https://docs.oracle.com/en/database/oracle/oracle-database/21/cncpt/oracle-database-instance.html#GUID-67247052-CE3F-44D2-BA3E-7067DEF4B6D5

Oracle Instance vs Database
https://stackoverflow.com/questions/17409862/database-vs-tablespace-whats-the-difference

Maybe instead of 'instance' we should use 'dbms_instance' and instead of 'database' say 'database_name'?

I'm open to all suggestions, and will attach some more robust content that uses both 'instance' and 'database' more than just .* and configure my test system to have a couple SQL instances on it.

solind:
Personally, I think "instance" and "database" would be good enough, provided we have better schema documentation and examples.

vanderpol:
I've revised the instance and database documentation, I'm guessing it will need further refinement, but figured this was a start.  I've attached some sample content and results.  This content is only for SQL Server at the moment, although SCC supports Oracle as well, I just don't have an Oracle DBMS system available right now.  We have plans to add support for PostgreSQL and MariaDB in the future.  The attached test content is a small subset of the DISA STIG content for SQL DB and SQL Instance.  When I find more free time, I may make some less DISA specific test content.
[Second_Sample_Results_SQL512.zip](https://github.com/user-attachments/files/17416071/Second_Sample_Results_SQL512.zip)
[sql512_test_content.xml.txt](https://github.com/user-attachments/files/17416072/sql512_test_content.xml.txt)

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -1309,6 +1321,179 @@
...
+                                                <xsd:element name="instance" type="oval-def:EntityObjectStringType">
+                                                      <xsd:annotation>
+                                                            <xsd:documentation>The instance entity defines the specific instance name to be used when connecting to the correct database, where instance refers to the running instance of the DMBS software itself. This could be a separate installation of binaries (such as with MS SQL Server), or just a set of running processes used to manage the DBMS.</xsd:documentation>
+															<xsd:documentation>The OVAL interpreter will automatically determine the list of available instances on the target.</xsd:documentation>
+															<xsd:documentation>When a pattern or string is entered, the OVAL interpeter will consider any matching instance as in scope for analysis.</xsd:documentation>
+													  </xsd:annotation>
+                                                </xsd:element>
+                                                <xsd:element name="database" type="oval-def:EntityObjectStringType" nillable="true">
```
solind, line 1416: So, what does it mean if `xsi:nil="true"`?  (ETA - another comment below)

For that matter, what if the database doesn't have a "Tablespace" type of concept (e.g., Derby DB -- which isn't supported in OVAL, but I happen to know it doesn't have such a concept)?

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -1309,6 +1321,179 @@
...
+                                                      <xsd:annotation>
+                                                            <xsd:documentation>The database entity defines the specific database name to be used when connecting to the specified instance, where a database is defined as a collection of tables within a DBMS instance.</xsd:documentation>
+															<xsd:documentation>When a pattern or string is entered, the OVAL interpeter will perform the query against any matching databases.</xsd:documentation>
+															<xsd:documentation>If the xsi:nil attribute is set to true, then only the target instance's default database will be queried.</xsd:documentation>
```
solind, line 1420: Duh, read ahead, right?  But still, in Oracle, for instance, there's a default Tablespace that can be defined for each user, but I don't believe there's like a universally "default" database defined.

So if xsi:nil="true", it's going to depend on the user who is used to log in.  And the user is going to be left up to the tool, just like for OS scanning.

solind (commented):
We should probably document what we mean by default when `database@xsi:nil="true"`.  If it's okay to mean that it's the default for whatever user is used to log in, then we should just say that.  But also, then... in that case, should we have a username element in the sql512_item?

oval-schemas/independent-definitions-schema.xsd
vanderpol: if xsi:nil = 'true', we just run the query against the DBMS instance, not a single specific database.  I was debating the whole 'default' database terminology as it was confusing to me as well.  Sometimes these queries are against a specific 'master' database, but other queries are are very bizzare, and may be SQL Server centric, but maybe other DBMS's have similar things, see below:

SELECT
    SERVERPROPERTY('IsClustered') AS [is_clustered],
    SERVERPROPERTY('IsHadrEnabled') AS [is_hadr_enabled]

This queries some configuration setting for the SQL Server instance, and is not targeted at any single database table.

vanderpol:
> We should probably document what we mean by default when `database@xsi:nil="true"`. If it's okay to mean that it's the default for whatever user is used to log in, then we should just say that. But also, then... in that case, should we have a username element in the sql512_item?

Adding a username will make the content only be useful to a single server, unless there are generic usernames that are present on all Oracle databases that you would know about in advance?  Given that we would want to digitally sign the content, our goal would be content that could be used on 'all' database systems without being edited.

oval-schemas/independent-definitions-schema.xsd
vanderpol: Agreed, the tool will determine which credential to use, just like with OS scanning.

solind:
@vanderpol I was not suggesting adding a username to the definitions, but to the collected system-characteristics (which are by nature specific to a single server).  I recall Joval put the login username somewhere in the results; I don't recall if XCCDF results had a specific place for that, or if we put it into a metadata field.

vanderpol:
@solind, thanks for the clarification.  I could see where knowing what username/credentials were used to scan the target dbms, as we save the credentials for scanning the OS as part of XCCDF results such as  <cdf:identity authenticated="true" privileged="true">DESKTOP-7ODDPV4\Jack</cdf:identity>  I'm not sure if there's a place in any OVAL results to store identiy/credentials?  I know we don't currently.

solind:
@vanderpol That was the place!  The `cdf:identity` field!

I don't think there's a formal OVAL equivalent, but there are plenty of "metadata" fields that can have `xs:any` children.  I guess you're suggesting the database user could just be stored in one of those places?

vanderpol:
I'm not really suggesting anything at this point, your idea of a username/identity as part of the collected item is interesting and I'm wondering why we don't have an identity globally in the oval results and pondering if we should.  It wouldn't replace the need for at the SQL item level.  If there was enough interest, I'd say we do both in a standard method.  Using gene

vanderpol:
Accidentally closed, fat fingers on my phone..

vanderpol:
In updating the documentation I realized that the 'version' design from the former developer was well intentioned, but over simplified, converting 'real' versions to human readable versions (example for SQL Server 2016, the real version is something like 13.0.6450.1.  I'll be updating sample content next week to replace the version of '2016' with a pattern match of ^13\.0\..*

This method will allow tools and content authors to use 'standard', well documented versions for each product, even if they are not quite a easy to understand or correlate to their end user visible versions.

vanderpol:
@solind Sorry for the extra rounds of documentation and test content review.  In writing documentation and attempting to explain it to you, I realized that the existing SQLEXT spec, which I used to create the SQL512 was definitely under specified, and the version field being translated likely won't age well vs just using the 'real' version.

solind:
@vanderpol Yikes, good catch!  No worries, I'll watch out for that commit.

vanderpol:
@solin I committed it, but am wondering if I should be making the version field a version data type.  Will research that next week

vanderpol:
@solind any thoughts on "version" being a stringType vs a versionType?  It's been string for both the 'sql' and 'sql57' test (not that they were ever really used).  I can see benefits for each.  As the data is pretty disparate across vendors, a string is more forgiving and flexible, where as versionType allows for less than, greater than etc... and I think should still work with non-numeric data such as '10.4.7-MariaDB'

I have it updated in my code to use versionType, but would like a sanity check before committing.

solind:
@vanderpol Indeed, the "string" datatype is much more forgiving.  Most databases use traditional version schemes, but it's not always clear what deserves to be thought of as the "version".

Take a favorite like MS SQL.  Is the version "SQL Server 2022", or is it "16.0.4135.4" (which they call the "build number") -- or is it "16" or "16.0"?

IIRC, with Joval, we compared the supplied string value using engine-specific logic, to determine whether we had an available compatible driver to make the connection.  So, I'd probably just keep it a string, until we can better define what exactly we're looking for.

solind (commented):
Hmm, wasn't the documentation actually more specific/clear before making this change?

vanderpol:
I removed some a schematron rule which I didn't create (former developer on our team wrote it), and didn't understand. and a couple other items were abbreviated a bit.  If you have any specific concerns, I'll pull back the previous commit.

solind:
Huh, when I wrote that last comment of mine the changes consisted solely of the removal of some doc annotations from independend-definitions-schema.xsd.  Now the changes look great!  Weird.

solind (approved):

vanderpol:
Thanks @solind, as always your feedback is very much appreciated!

PR #186, opened 2024-11-08
Added several new sql dbms management systems to EntityObjectEngineType 

vanderpol:
Added several newer SQL DBMS engines for possible future use

### DIFF
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -2502,6 +2502,26 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
                                     <xsd:documentation>The access value describes the Microsoft Access database engine.</xsd:documentation>
                               </xsd:annotation>
                         </xsd:enumeration>
+                        <xsd:enumeration value="aurora">
+                              <xsd:annotation>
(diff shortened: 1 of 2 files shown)


solind (changes requested):
You need to make corresponding changes to  the `EntityItemEngineType` in `independent-system-characteristics-schema.xsd`.

vanderpol:
Thanks @solind for that catch.  I'm wondering if for OVAL 6.0 there's a way to prevent all of these duplicated enumerations that we tend to forget to update with time?  Feels like these enumerations need to be part of a 'common' file vs duplicated in definitions and system-chars.

Update is ready for your review.

solind (approved):

solind:
> I'm wondering if for OVAL 6.0 there's a way to prevent all of these duplicated enumerations that we tend to forget to update with time? Feels like these enumerations need to be part of a 'common' file vs duplicated in definitions and system-chars.

I think it should be investigated as something we could do with schema consolidation.  First someone needs to research why/how it came to be the way it is now!

vanderpol:
@solind, I think we'd have to talk to the original guys at MITRE who developed the spec to know (maybe Jon Baker or Danny Haynes?) Either way, we are where we are... I'll ponder the level of effort to improve it for 6.0, and see if we have time.

PR #187, opened 2024-11-16
Remove dirs from gh action clean script

DeafBDoor:
OVAL-Board, Area Supervisors, and Email List Archives dirs were removed from the repository, so they are not needed into the clean up script anymore.


dodys (approved):
lgtm, thanks!

PR #190, opened 2024-11-16
Generate html doc on release workflow

DeafBDoor:
This PR consists of two parts:
- a github action (named buildDocs-action) which runs a Docker container to build the documentation
- changes to the buildRelease workflow so it uses the buildDocs-action

Note that - currently - the buildRelease workflow is not handling the release properly, so additional changes are needed.
I'm marking this PR as a draft for this reason alone. 

Update: Release is now being generated as a draft with all files required. PR was also simplified to use a single workflow with two jobs + an action which is responsible for the Docker container which builds the docs.


DeafBDoor:
Solution now should work without issues. It just requires a version label as a parameter, then it generates a draft release containing the source and the HTML docs.

vanderpol:
@DeafBDoor, sounds very promising, I really appreciate your efforts on this.   Can I see the test results of this on your branch of this repository, or do we need to do it live on our repository to test?

vanderpol:
@DeafBDoor thanks for the invite to see your fork of the OVAL project.  I have a couple minor suggestions:
1.  Recommend renaming OVAL_HTML to be OVAL-<version>-Documentation
2.  In the OVAL-<version>.zip file, we can remove the oval-specifications directory, for 2 reasons.   First is that it's also included in the OVAL HTML documentation zip, but primarily because they are no longer being updated, and we may drop them entirely very soon, unless someone has a week or more of time to update them.
3. In the generated HTML files, something isn't quite right, if you click on any of the hyperlinks on the left pane, you end
 up seeing  the source HTML (see screenshot)  
image

vanderpol:
@DeafBDoor please get with @A-Biggs  regarding the potential to use a different XSLT transforms to create HTML docs based on the old MITRE documentation, which are easier to read and don't have the platform overlap issue with dnscache and others.  If your changes are ready to be pulled into master, please let me know as well.

DeafBDoor:
Hey, sorry for the radio silence. I’m moving to another country soon, so
all my available went to bureaucratic work.

Do we have a reference on how to work with these XLST transforms?

Kind regards,
Richard Maciel Costa


Em seg., 25 de nov. de 2024 às 12:37, Jack Vander Pol <
***@***.***> escreveu:

DeafBDoor:
> Hey @DeafBDoor, I know you weren't on today's OVAL Board meeting call, but the release docs came up as a discussion point, and the thought was where the Read The Docs (.rst files) were causing issues with links being wrong, that for the release, it might be better to do something similar to what is done in the Joval Repo [here](https://github.com/joval/jOVAL/blob/master/scap/GNUmakefile) to build the HTML docs directly from the XSDs, where that doesn't have the link issue. I was going to take a crack at getting the github runner to build the docs that way and zip them up so that what is included with the release has the links correct. If you are able to figure out how to fix the links using the existing build process, let me know and we can maybe go with that.

Sure, I can investigate the problem, but I believe your idea to use a mechanism that already works is much better. I can change the Dockerfile responsible for generating the docs to use jOVAL instead. It's just a matter of getting the correct set of requirements.

DeafBDoor:
Tested docker image `docker.io/library/eclipse-temurin:8` and only had to install `make`.
It ran the makefile flawlessy.

So, my idea here is to just substitute the Dockerfile which uses python to generate the documentation by one which uses eclipse-temurin java 8. We can directly use the commands from jOVAL (by cloning the repo), or copy the commands from the jOVAL repo (given that licenses are compatible).

I don't like the idea of adding jOVAL as a dependency to OVAL, because it would create a circular dependency, so I vote for copying the makefiles and modyfing them to just generate the documentation.

A-Biggs:
FYI, I just put up a [PR](https://github.com/OVAL-Community/OVAL/pull/196) with what I was thinking for the release process for creating the docs based on the Joval way.  Have a look at that.

vanderpol:
Are both PR's needed?  Glad to approve whatever is needed, just let me know.

A-Biggs:
I don't think both are needed.  I think the https://github.com/OVAL-Community/OVAL/pull/196 does everything we need when it comes to releases.

At some point we should visit how to generate and update the Read The Docs pages (or maybe update them to not be generated from .rst files, or replace that page to the latest release zip file), but I see that effort being outside of the scope of this release github workflow.

vanderpol:
@A-Biggs Sounds good.  For now I can run the transforms on my computer and commit the updated .rst files for readTheDocs, and figure out a more elegant solution later.

Issue #150, opened 2024-09-20
Add new independent test to support running shell commands called "shellcommand"


vanderpol:
**Abstract**

Add an independent test called "shellcommand" which supports running any arbitrary command or contents of a shell script on the target system.  This feature should only be used if the content is trusted, ideally digitally signed either at the benchmark level, OVAL document level, or at the object level.

This allows for many more system configuration and vulnerability tests to be performed with OVAL without making the language overly large by adding numerous system commands to OVAL.   This also will allow for security personal not well versed with OVAL to be able to create OVAL content.  It should also help to simplify some of the overly complex binary trees of tests that has plagued OVAL in the past.

**Link to Proposal**
Once a proposal has been put into pull request form, add a link to the PR here. If and as alternate proposals or objections are added they should be linked here as well.

**Additional context**
Add any other context or screenshots about the enhancement.

shawndwells:
1. Red Hat may have some opinions and lessons learned from implementation of Script Check Engine (https://www.open-scap.org/features/other-standards/sce/)

2. Can this also be implemented for *all* platforms? Specifically Windows via PowerShell and Cisco iOS and NX-OS?

vanderpol:
@shawndwells Thanks for your input, we will check out the Script Check Engine for lessons learned there.  As for Windows we plan for a windows specific windows;powershellcommand test (which essentially mirrors the unix:shellcommand, but allows for specific backend support for differences in the environments.  We had not considered Cisco IOS, IOS XE or NX-OS as in scope for this, given the Cisco specific schemas which allow for running any global/line/interface command, I didn't see where Cisco was needing this type support.

johnulmer-oval:
Please find the draft proposal for the unix shellcommand test attached.  The introduction (Background) section of the attached PDF follows.  


Background

OVAL test types have been relatively specific in their targeted functions and, typically, have been applicable to a specific system, sub-system, or application on a target machine. The time and effort required to add new tests to the OVAL specification and the relative scarcity of resources impede efficient development and support the specification. To counter this and to broadly expand the breadth of OVAL tests in general, we propose two new OVAL tests, one each for the UNIX and Windows families, the Unix shell_comand test and the Windows powershell_command test. Each is much broader in their application than the previously narrow tests currently available in OVAL allowing a content author to use the full capability of the Unix shell (e.g. bash) or the Windows powershell to retrieve system characteristics.

[...]

PR #192, opened 2024-11-21
Add new independent test to support running shell commands called shellcommand

vanderpol:
This pull request adds the capability discussed in #150

### DIFF
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -1039,7 +1039,7 @@ The evaluation of the object should always produce at least one item.  If the ob
                                     <xsd:choice>
                                           <xsd:element ref="oval-def:set"/>
                                           <xsd:sequence>
-                                                <xsd:element name="shell" type="oval-def:EntityObjectShellType" use="optional" default="'cmd' on Windows, 'sh' on UNIX/linux, 'Zsh' on MacOS" >
+                                                <xsd:element name="shell" type="oval-def:EntityObjectShellType" use="required" >
                                                       <xsd:annotation>
                                                             <xsd:documentation>The shell entity defines the specific shell to use (e.g. bash, csh, ksh, etc.). Any tool collecting information for this object will need to know the shell in order to use it properly.</xsd:documentation>
                                                       </xsd:annotation>
@@ -2682,9 +2682,9 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
                                     <xsd:documentation>The korn shell (ksh).</xsd:documentation>
                               </xsd:annotation>
                         </xsd:enumeration>
-                        <xsd:enumeration value="Zsh">
+                        <xsd:enumeration value="zsh">
                               <xsd:annotation>
-                                    <xsd:documentation>The Z shell (Zsh).</xsd:documentation>
+                                    <xsd:documentation>The Z shell (zsh).</xsd:documentation>
                               </xsd:annotation>
                         </xsd:enumeration>
                         <xsd:enumeration value="cmd">
@@ -2731,9 +2731,9 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
                                     <xsd:documentation>The korn shell (ksh).</xsd:documentation>
                               </xsd:annotation>
                         </xsd:enumeration>
-                        <xsd:enumeration value="Zsh">
+                        <xsd:enumeration value="zsh">
                               <xsd:annotation>
-                                    <xsd:documentation>The Z shell (Zsh).</xsd:documentation>
+                                    <xsd:documentation>The Z shell (zsh).</xsd:documentation>
                               </xsd:annotation>
                         </xsd:enumeration>
                         <xsd:enumeration value="cmd">
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -969,6 +969,153 @@
             </xsd:complexType>
       </xsd:element>
       <!-- =============================================================================== -->
+      <!-- ============================  SHELLCOMMAND TEST  =============================== -->
+      <!-- =============================================================================== -->
+      <xsd:element name="shellcommand_test" substitutionGroup="oval-def:test">
+            <xsd:annotation>
+                  <xsd:documentation>The shellcommand_test is used to check the values produced by the running of the command (or embedded script, not an external script file) found in the object 'command' element. It extends the standard TestType as defined in the oval-definitions-schema and one should refer to the TestType description for more information. The required object element references a shellcommand_object and the optional state element references a shellcommand_state that specifies the information to check.
+                                   
+Since this test requires the running of code supplied by content and since SCAP applications commonly run with elevated privileges, significant responsibilty falls to the content author to do no harm.  This also requires that any content stream that employs this test MUST be from a known trusted source and be digitally signed.  The use of any executables that are not supplied by the installed operating system is highly discouraged.</xsd:documentation>
+                  <xsd:appinfo>
+                        <oval:element_mapping>
+                              <oval:test>shellcommand_test</oval:test>
+                              <oval:object>shellcommand_object</oval:object>
+                              <oval:state>shellcommand_state</oval:state>
+                              <oval:item target_namespace="http://oval.mitre.org/XMLSchema/oval-system-characteristics-5#independent">shellcommand_item</oval:item>
+                        </oval:element_mapping>
+                  </xsd:appinfo>
+                  <xsd:appinfo>
+                        <sch:pattern id="ind-def_shellcommand_tst">
+                              <sch:rule context="ind-def:shellcommand_test/ind-def:object">
+                                    <sch:assert test="@object_ref=ancestor::oval-def:oval_definitions/oval-def:objects/ind-def:shellcommand_object/@id"><sch:value-of select="../@id"/> - the object child element of a shellcommand_test must reference a shellcommand_object</sch:assert>
+                              </sch:rule>
+                              <sch:rule context="ind-def:shellcommand_test/ind-def:state">
+                                    <sch:assert test="@state_ref=ancestor::oval-def:oval_definitions/oval-def:states/ind-def:shellcommand_state/@id"><sch:value-of select="../@id"/> - the state child element of a shellcommand_test must reference a shellcommand_state</sch:assert>
+                              </sch:rule>
+                        </sch:pattern>
+                  </xsd:appinfo>
+            </xsd:annotation>
+            <xsd:complexType>
+                  <xsd:complexContent>
+                        <xsd:extension base="oval-def:TestType">
+                              <xsd:sequence>
+                                    <xsd:element name="object" type="oval-def:ObjectRefType" />
+                                    <xsd:element name="state" type="oval-def:StateRefType" minOccurs="0" maxOccurs="unbounded"/>
+                              </xsd:sequence>
+                        </xsd:extension>
+                  </xsd:complexContent>
+            </xsd:complexType>
+      </xsd:element>
+      
+      <xsd:element name="shellcommand_object" substitutionGroup="oval-def:object">
+            <xsd:annotation>
+                  <xsd:documentation>The shellcommand_object is used by a shellcommand_test to define a command (or shell script) to be run and a pattern to filter result lines. Each object extends the standard ObjectType as defined in the oval-definitions-schema and one should refer to the ObjectType description for more information. The common set element allows complex objects to be created using filters and set logic.</xsd:documentation>
+                  <xsd:appinfo>
+                        <sch:pattern id="ind-def_shellcommand_object_verify_filter_state">
+                              <sch:rule context="ind-def:shellcommand_object//oval-def:filter">
+                                    <sch:let name="parent_object" value="ancestor::ind-def:shellcommand_object"/>
+                                    <sch:let name="parent_object_id" value="$parent_object/@id"/>
+                                    <sch:let name="state_ref" value="."/>
+                                    <sch:let name="reffed_state" value="ancestor::oval-def:oval_definitions/oval-def:states/*[@id=$state_ref]"/>
+                                    <sch:let name="state_name" value="local-name($reffed_state)"/>
+                                    <sch:let name="state_namespace" value="namespace-uri($reffed_state)"/>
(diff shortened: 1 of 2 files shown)


vanderpol:
I've added anyone who commented on issue #150 as a reviewer of this.  We really just need one person to sign off on this, as it's not a complex addition.  As soon as this is pulled into developed OVAL 5.12 is essentially 'done'

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                                    <xsd:element name="state" type="oval-def:StateRefType" minOccurs="0" maxOccurs="unbounded"/>
+                              </xsd:sequence>
+                        </xsd:extension>
+                  </xsd:complexContent>
+            </xsd:complexType>
+      </xsd:element>
+      
+      <xsd:element name="shellcommand_object" substitutionGroup="oval-def:object">
+            <xsd:annotation>
+                  <xsd:documentation>The shellcommand_object is used by a shellcommand_test to define a shell to use (e.g. sh, bash, ksh, etc.), a command (or shell script) to be run, and a pattern to filter result lines. The default shell is bash.  Each object extends the standard ObjectType as defined in the oval-definitions-schema and one should refer to the ObjectType description for more information. The common set element allows complex objects to be created using filters and set logic.
```
solind, line 1017: I think the default shell should be:
1) cmd.exe on Windows
2) /bin/sh on Unix/Linux/macOS

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                  
+The evaluation of the object should always produce at least one item.  If the object evaluation does not produce output that should result in an item, one should be created with a status of 'does not exist' and, if possible, the exit_status and any STDERR from the system call should be captured in that item.
```
solind, line 1019: Should read: "... should always produce one item."

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+            </xsd:annotation>
+            <xsd:complexType>
+                  <xsd:complexContent>
+                        <xsd:extension base="oval-def:ObjectType">
+                              <xsd:sequence>
+                                    <xsd:choice>
+                                          <xsd:element ref="oval-def:set"/>
+                                          <xsd:sequence>
+                                                <xsd:element name="shell" type="oval-def:EntityObjectStringType">
+                                                      <xsd:annotation>
```
solind, line 1043: I think it's potentially problematic to use a relative path and an arbitrary string value for specifying the path to the shell.  How about an enumeration?

If we do an enumeration, I'd have values for: Korn, Bourne, csh, tcsh, bash, zsh, /usr/bin/env, cmd.exe, powershell.exe and "" (empty string to allow for a variable value).

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+															
+A subexpression (using parentheses) can call out a piece of the matched stdout_line to test. For example, the pattern abc(.*)xyz would look for a block of text in the file that starts with abc and ends with xyz, with the subexpression being all the characters that exist in between. The value of the subexpression can then be tested using the subexpression entity of a shellcommand_state. Note that if the pattern, starting at the same point in the line, matches more than one block of text, then it matches the longest. For example, given a file with abcdefxyzxyzabc, then the pattern abc(.*)xyz would match the block abcdefxyzxyz. Subexpressions also match the longest possible substrings, subject to the constraint that the whole match be as long as possible, with subexpressions starting earlier in the pattern taking priority over ones starting later.
```
solind, line 1056: "... in the _file_ ..." should be changed to "... in the _output_ ..."

also "... given a _file_ with ..." -> "... given _output_ with ..."

oval-schemas/independent-definitions-schema.xsd
solind: ... per variable value permutation, naturally, being implied.

oval-schemas/independent-definitions-schema.xsd
solind: OOPS!  You already have an enumeration (EntityObjectShellType).  You just set the type attribute wrong for the shell entity.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+      <xsd:element name="shellcommand_state" substitutionGroup="oval-def:state">
+            <xsd:annotation>
+                  <xsd:documentation>The shellcommand_state contains the entities that are used to check the values returned by the shellcommand_object.  Note that the state entities shell, command, and pattern are echoed, verbatim, from the same elements in the associated shellcommand_object.</xsd:documentation>
+            </xsd:annotation>
+            <xsd:complexType>
+                  <xsd:complexContent>
+                        <xsd:extension base="oval-def:StateType">
+                              <xsd:sequence>
+                                    <xsd:element name="shell" type="oval-def:EntityStateStringType" minOccurs="0" maxOccurs="1">
+                                          <xsd:annotation>
```
solind, line 1087: type attribute needs to be EntityStateShellType.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                                                <xsd:documentation>The 'shell' element contains the shell used to perform the command and must match the value in the associated object, verbatim.</xsd:documentation>
+                                          </xsd:annotation>
+                                    </xsd:element>
+                                    
+                                    <xsd:element name="command" type="oval-def:EntityStateAnySimpleType" minOccurs="1" maxOccurs="1">
+                                          <xsd:annotation>
```
solind, line 1093: Why AnySimpleType?  Shouldn't it just be a StringType?

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                                                      </sch:pattern>
+                                                </xsd:appinfo>
+                                          </xsd:annotation>
+                                    </xsd:element>
+			                        <xsd:element name="exit_status" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
+										<xsd:annotation>
+											<xsd:documentation>The 'exit_status' entity represents the exist status returned by the system for the execution of the object command.</xsd:documentation>
+										</xsd:annotation>
+									</xsd:element>
+                                    <xsd:element name="stdout_line" type="oval-def:EntityStateAnySimpleType" minOccurs="0" maxOccurs="unbounded">
```
solind, line 1114: Should be an EntityStateStringType.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                                          <xsd:annotation>
+                                                <xsd:documentation>The 'stdout_line' entity represents a line from the STDOUT output of a successful run of the command string that matched the specified object pattern.</xsd:documentation>
+                                          </xsd:annotation>
+                                    </xsd:element>
+                                    <xsd:element name="subexpression" type="oval-def:EntityStateAnySimpleType" minOccurs="0" maxOccurs="1">
+                                          <xsd:annotation>
```
solind, line 1120: Should probably be an EntityStateStringType. Unless you imagine someone capturing numeric output and using numerical comparisons in a state.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -968,6 +968,171 @@
...
+                                                <xsd:documentation>The subexpression entity represents a value to test against the subexpression in the specified pattern. If multiple subexpressions are specified in the pattern, this value is tested against all of them. For example, if the pattern abc(.*)mno(.*)xyp was supplied, and the state specifies a subexpression value of enabled, then the test would check that both (or at least one, none, etc. depending on the entity_check attribute) of the subexpressions have a value of enabled.</xsd:documentation>
+                                          </xsd:annotation>
+                                    </xsd:element>
+                                    <xsd:element name="stderr_line" type="oval-def:EntityStateAnySimpleType" minOccurs="0" maxOccurs="unbounded">
+                                          <xsd:annotation>
```
solind, line 1125: Should probably be an EntityStateStringType, since the error output is pretty much impossible to control.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -2491,6 +2656,94 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
...
+                        <xsd:enumeration value="csh">
+                              <xsd:annotation>
+                                    <xsd:documentation>The C shell (csh).</xsd:documentation>
+                              </xsd:annotation>
+                        </xsd:enumeration>
+                        <xsd:enumeration value="ksh">
+                              <xsd:annotation>
+                                    <xsd:documentation>The korn shell (ksh).</xsd:documentation>
+                              </xsd:annotation>
+                        </xsd:enumeration>
```
solind, line 2684: Add zsh for sure, since that's the standard shell on macOS now.

oval-schemas/independent-system-characteristics-schema.xsd
```diff
@@ -325,6 +325,64 @@
...
+               <xsd:documentation>The shellcommand_item stores information retrieved from the local system that results from the running of the command or embedded script in the associated object command element.
+               
+The evaluation of the object should always produce at least one item.  If the object evaluation does not produce output via STDOUT that should result in an item, one should be created with a status of 'does not exist'.  This facilitates that capture of the exit_status and stderr from the system call.</xsd:documentation>
+          </xsd:annotation>
+          <xsd:complexType>
+               <xsd:complexContent>
+                    <xsd:extension base="oval-sc:ItemType">
+                         <xsd:sequence>
+                              <xsd:element name="shell" type="oval-sc:EntityItemStringType" minOccurs="1" maxOccurs="1">
+                                   <xsd:annotation>
```
solind, line 344: type needs to be EntityItemShellType

oval-schemas/independent-system-characteristics-schema.xsd
```diff
@@ -325,6 +325,64 @@
...
+								<xsd:annotation>
+									<xsd:documentation>The exit_status entity represents the exist status returned by the system for the execution of the object command. OVAL Item status should match the exit status of the system call.</xsd:documentation>
+								</xsd:annotation>
+							</xsd:element>
+							<xsd:element name="stdout_line" type="oval-sc:EntityItemStringType" minOccurs="0" maxOccurs="unbounded">
+								<xsd:annotation>
+									<xsd:documentation>The stdout_line entity contains a line from the STDOUT output of a successful run of the command string that matched the specified object pattern.  Each line created by the execution of the object command should create an item 'stdout_line' element.</xsd:documentation>
+								</xsd:annotation>
+							</xsd:element>
+                            <xsd:element name="subexpression" type="oval-sc:EntityItemAnySimpleType" minOccurs="0" maxOccurs="unbounded">
```
solind, line 369: I'd just make it EntityItemStringType.  The state comparison would govern the type, if needed.

solind (changes requested):
Some likely copy/paste issues, and a few substantive issues raised.

oval-schemas/independent-definitions-schema.xsd
vanderpol: @solind how do we go about having 2 different default values?  I'm obviously ignorant on this, I assumed in XML you could only have one default, which is partly why we originally wanted to have this be 2 different tests, one for Windows and one for UNIX.

oval-schemas/independent-definitions-schema.xsd
vanderpol: Yeah this was likely due to my updates and then John's, I'll fix the shell to use the enumeration.

oval-schemas/independent-definitions-schema.xsd
vanderpol: We have some examples of STIG content where capturing a numeric value and doing mathematical operations is needed.

oval-schemas/independent-definitions-schema.xsd
vanderpol: agreed.  I'll change this.

oval-schemas/independent-definitions-schema.xsd
vanderpol: Good catch.  Will update

oval-schemas/independent-system-characteristics-schema.xsd
vanderpol: Confirmed.

oval-schemas/independent-definitions-schema.xsd
vanderpol: Confirmed.

oval-schemas/independent-definitions-schema.xsd
vanderpol: Confirmed.

oval-schemas/independent-system-characteristics-schema.xsd
vanderpol: Current design matches textfilecontent's subexpression design.  Should we change both this and textfilecontent, or leave them both as-is?

vanderpol:
@solind with the exception of the subexpression entity item type (which matches methods from textfilecontent) I think I've made all of your recommended changes.    I'm open to changing the subexpression as well, just feels wrong to have it different from the other test we are basing it on.

oval-schemas/independent-definitions-schema.xsd
solind: @vanderpol You can't. If the element is optional, however, and there is no default value set ... then we can document a meaning for that, which would be to set it to the platform-specific value.

oval-schemas/independent-system-characteristics-schema.xsd
solind: @vanderpol No kidding! Well, there's certainly no harm in its being an EntityItemAnySimpleType, so might as well just leave it.

vanderpol:
@solind unless you have a solution for the different default shells per OS, I think it's ready for merge.

oval-schemas/independent-definitions-schema.xsd
solind: But you didn't change the text to say: "The evaluation of the object should always produce one item."

solind: Actually, I think you should rewrite it as follows:

```
The evaluation of the object should always produce one item. If the command execution does not produce output, an item should still be created with the exit_status (AKA process exit code), a stdout entity with a status of 'does not exist', and any STDERR from the execution captured in stderr_line entities.
```

solind (changes requested):
It's nearly there, except for the documentation note in the one unresolved conversation.

vanderpol:
Thanks @solind, I thought I changed that, but it either I missed it, or it existed in more than one location.  Committed based on your latest recommendation.

solind (approved):

Issue #134, opened 2023-03-17
Add Support for IPv4 and IPv6, both, to interfacesType


johnulmer-oval:
**Describe the bug**
Currently, the interfacesType in the oval-system-characteristics-schema schema allows only one IP address.  As IPv6 addresses come into common use, that schema needs updating.  Suggest that the schema should allow more than one ip_address element and add an attribute to that element for 'ip_address_version'.

**Replication Steps** (if applicable)
Use any SCAP processor to create OVAL results.  

**Expected behavior**
For network interfaces that have more than one IP address configured, allow the reporting of both IPv4 and IPv6 addresses.

Technically, while not common, it is possible to configure a single interface to have more than one address of each type.  I don't see this often and am open to the idea that updating of the referenced schema might be more robust if it allowed multiple addresses of any type.  However, in common use, currently, it appears that it is typical to have only one IPv4 and only one IPv6 for a given interface.

vanderpol:
I second this item, primarily to ensure it's not auto-closed.

PR #193, opened 2024-11-25
Update ip address field to use IPAddressType to include ipv4 vs ipv6 data

vanderpol:
Modified the 'type' for the 'ip_address' element in an interface from 'xsd:string' to oval-sc:EntityItemIPAddressType'.  This entity type already contains the needed 'datatype' attribute.

@solind, John @johnulmer-oval is out of the office for the week, and in my review of his update I'm wondering if he missed adding a maxoccurs=unbounded?  Can you review this change, (and the original issue #134) to see if I should be adding that for this merge?

### DIFF
--- a/oval-schemas/oval-system-characteristics-schema.xsd
+++ b/oval-schemas/oval-system-characteristics-schema.xsd
@@ -153,17 +153,17 @@
                     <xsd:documentation>The required interface_name element is the name of the interface</xsd:documentation>
                 </xsd:annotation>
             </xsd:element>
-            <xsd:element name="ip_address" minOccurs="0" maxOccurs="unbounded" type="xsd:IPv4Type">
+            <xsd:element name="ip_address" minOccurs="0" maxOccurs="unbounded" type="xsd:string">
                 <xsd:annotation>
                     <xsd:documentation>The ipv4_address element holds the IPV4 address for the interface.</xsd:documentation>
                 </xsd:annotation>
             </xsd:element>
-            <xsd:element name="ipv6_address" minOccurs="0" maxOccurs="unbounded" type="xsd:IPv6Type">
+            <xsd:element name="ipv6_address" minOccurs="0" maxOccurs="unbounded" type="xsd:string">
                 <xsd:annotation>
                     <xsd:documentation>The ipv6_address element holds the IPV6 address for the interface.</xsd:documentation>
                 </xsd:annotation>
             </xsd:element>
-            <xsd:element name="mac_address" minOccurs="0" type="xsd:MacAddressType" >
+            <xsd:element name="mac_address" minOccurs="0" type="xsd:string" >
                 <xsd:annotation>
                     <xsd:documentation>The required mac_address element holds the MAC address for the interface. MAC addresses should be formatted according to the IEEE 802-2001 standard which states that a MAC address is a sequence of six octet values, separated by hyphens, where each octet is represented by two hexadecimal digits. Uppercase letters should also be used to represent the hexadecimal digits A through F.</xsd:documentation>
                 </xsd:annotation>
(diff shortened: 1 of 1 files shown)


solind (commented):
Unfortunately I never looked at the original issue #134

I think we worked around the schema limitation in Joval by duplicating interfaces that had multiple addresses -- so you'd have multiple interface tags with the same values in their name sub-elements.  I think this would be a breaking change for everyone.

A better idea, although still a breaking change, is the one you suggested, which would be to allow multiple ip_address elements.

I'm also not sure how I feel about using an Item type in the system_info.  That feels wrong to me.

solind:
My recommendation would be to redefine InterfaceType to consist of a sequence of 4 elements:

- interface_name (minOccurs="1")
- ip_address (minOccurs="0" maxOccurs="unbounded")
- ip6_address (minOccurs="0" maxOccurs="unbounded")
- mac_address (minOccurs="0")

I'd also add xs:restriction elements to specify the string format of the different address types (IP4, IP6 and MAC).  See [this stack overflow article](https://stackoverflow.com/questions/59830797/xsd-regex-pattern-for-ip-number-with-machine-name-prefix).

vanderpol:
I'm always hesitant on regex implementations for IP addresses, but I'm trying this out with regexes from 

ipv4:
https://stackoverflow.com/questions/5284147/validating-ipv4-addresses-with-regexp

ivp6;
https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s17.html

macaddress:
https://stackoverflow.com/questions/4260467/what-is-a-regular-expression-for-a-mac-address

vanderpol:
Ignore my last commit, I was over-confident... and just tested it and I'm seeing errors I need to fix

[ERROR]14:17:33 - Error validating CPE OVAL Results: C:/Program%20Files/SCAP%20Compliance%20Checker%205.10.1%20RC1/Resources/Schemas/oval_5.11.2/oval-system-characteristics-schema.xsd:156: Schemas parser error : element decl. '{http://oval.mitre.org/XMLSchema/oval-system-characteristics-5}ip_address', attribute 'type': The QName value '{http://www.w3.org/2001/XMLSchema}IPv4Type' does not resolve to a(n) type definition.
C:/Program%20Files/SCAP%20Compliance%20Checker%205.10.1%20RC1/Resources/Schemas/oval_5.11.2/oval-system-characteristics-schema.xsd:161: Schemas parser error : element decl. '{http://oval.mitre.org/XMLSchema/oval-system-characteristics-5}ipv6_address', attribute 'type': The QName value '{http://www.w3.org/2001/XMLSchema}IPv6Type' does not resolve to a(n) type definition.
C:/Program%20Files/SCAP%20Compliance%20Checker%205.10.1%20RC1/Resources/Schemas/oval_5.11.2/oval-system-characteristics-schema.xsd:166: Schemas parser error : element decl. '{http://oval.mitre.org/XMLSchema/oval-system-characteristics-5}mac_address', attribute 'type': The QName value '{http://www.w3.org/2001/XMLSchema}MacAddressType' does not resolve to a(n) type definition

vanderpol:
@solind, I can fight my regex issues (the one above is just something dumb, but after fixing it, I ran into another round of regex issues).  Given that other areas of the the schema ipv4 and ipv6 are just 'strings', I'm wondering restricting their values here has much value?  I can make them strings, and be done in a couple minutes.  Seeing the debates online of what is the 'best' or 'most accurate' regex for IPV4 or IPV6 makes me concerned about issues going forward.

vanderpol:
I've reverted the datatype back to xsd:string for now, out of concerns of issues going forward, vs the risks of leaving as string data.  Given that other places in OVAL have just xsd:string for ipv4 and ipv6 data, I'm going for just fixing what this issue called for, allowing for both ipv4 and ipv6 to be present and valid

solind:
@vanderpol I thought [these XSD definitions](https://lists.w3.org/Archives/Public/www-xml-schema-comments/2005OctDec/0138.html) looked pretty sweet and all-encompassing.

solind:
But... if you're out of time, I'll try not to judge you too harshly if you just want to commit them as strings.

vanderpol:
Given everything else we have going on, and the benefits vs potential false positives, I'm just going to stick with strings and call 5.12 done.

PR #194, opened 2024-11-25
Added alternate name from MS documentation for all password policy items

vanderpol:
Added clarity to all of the password policy items, providing the exact name of the element matching MS documentation.


vanderpol:
@solind Not sure if you happened to see this one, just a trivial documentation change, but I expanded it to all elements in this group not just a single one.

solind (approved):
Looks fine

PR #213, opened 2024-12-10
Update oval 5 URL based namespace with mitre.org to URN based OVAL 6 namespaces

vanderpol:
Updated all MITRE centric URL namespaces with URN based OVAL 6 namespaces.  Note that I accidentally created this branch from 'master' instead of oval6-develop, so there are 2 .rst files with updates (that can be ignored for the scope of this PR)

### DIFF
--- a/guidelines/oval-schema-documentation/unix-definitions-schema.rst
+++ b/guidelines/oval-schema-documentation/unix-definitions-schema.rst
@@ -10,18 +10,18 @@ The OVAL Schema is maintained by the OVAL Community. For more information, inclu
 
 Test Listing  
 ---------------------------------------------------------
-* :ref:`dnscache_test` (Deprecated)  
-* :ref:`file_test`  
-* :ref:`fileextendedattribute_test` (Deprecated)  
-* :ref:`gconf_test` (Deprecated)  
-* :ref:`inetd_test` (Deprecated)  
+* :ref:`dnscache_test_unix` 
+* :ref:`file_test_unix`  
+* :ref:`fileextendedattribute_test`  
+* :ref:`gconf_test` 
+* :ref:`inetd_test` 
 * :ref:`interface_test`  
 * :ref:`password_test`  
-* :ref:`process_test` (Deprecated)  
-* :ref:`process58_test`  
-* :ref:`routingtable_test` (Deprecated)  
+* :ref:`process_test_unix`   
+* :ref:`process58_test_unix`  
+* :ref:`routingtable_test` 
 * :ref:`runlevel_test`  
-* :ref:`sccs_test` (Deprecated)  
+* :ref:`sccs_test` 
 * :ref:`shadow_test`  
 * :ref:`sshd_test`  
 * :ref:`symlink_test`  
@@ -31,9 +31,9 @@ Test Listing
   
 ______________
   
-.. _dnscache_test:  
+.. _dnscache_test_unix:  
   
-< dnscache_test > (Deprecated)  
+< dnscache_test > (unix) (Deprecated)  
 ---------------------------------------------------------
 Deprecation Info  
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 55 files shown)


vanderpol:
@solind, @balleman-ctr, per our discussion #169, I've updated all of the .xsd files with non-MITRE, URN formatted namespaces.  As I noted in the PR, you can ignore the 2 .rst files which accidentally got pulled into this, as I branched from master instead of oval6-develop.  They are just noise for the sake of this PR.   Along with updating all of the "https://mitre.org..." with "urn:oval:..." I also updated several 'uri="https://mitre" to 'urn="urn:oval:v6".  Schema's validated with them set as 'uri', but it seemed more correct to update it to be 'urn='

solind:
@vanderpol I assume you just did a find/replace of all the URL strings, which should certainly work.  The thing to do to verify these changes, however, is to use the new XSDs to generate a JAXB or Python data model. I'm not really in a position to try that, but @maxullman or @A-Biggs might be, if they have some spare cycles.

vanderpol:
I performed a collection of find/replace all, and ran a python script which loaded each of the .xsd files using the 'xmlschema' module between each update to make sure I didn't cause any obvious issues.  I'll some additional internal testing, but would really appreciate any independent testing that anyone can offer up.

vanderpol:
In our code, the new URN based schemas are causing issues, which I'll have to research to see if there are workarounds, for example, we were splitting the platform schema on # to pull of things like 'windows', or 'independent' from the end of the URL.  There might be other ways to implement, but we're going to have to change some code to get it all to work.

If we want to decrease any burden on OVAL interpreters, switching to the custom URL method like http://oval-community.org/v6/system-chararacteristics#windows vs urn:oval:v6:system-characteristics:windows might be wise.

solind:
I would expect any change in the namespaces to break existing code.  I know that Joval has defined string constants based on the namespaces, and therefore those references will certainly break.

Nevertheless, fixing that is straightforward, and I think we should expect this kind of thing going from 5.12 to 6.0.  Generating a JAXB data model, however, would tell us if the namespaces are all internally consistent, which is the real question.

vanderpol:
I have manually updated the DISA STIG Windows 11 content to use the proposed format, I also had to hack the SCAP 1.3 source data stream schema, and have been able to successfully validate it in SCC (which uses libxml2) for validation.  Seems like a pretty decent test so far.

12:24:57 - Validating input files, this may take a while.
12:24:57 - Validating U_MS_Windows_11_V2R2_STIG_SCAP_1-3_Benchmark-enhancedV8.xml

[scap-source-data-stream_1.3.xsd.txt](https://github.com/user-attachments/files/18083535/scap-source-data-stream_1.3.xsd.txt)

[U_MS_Windows_11_V2R2_STIG_SCAP_1-3_Benchmark-enhancedV8.xml.txt](https://github.com/user-attachments/files/18083534/U_MS_Windows_11_V2R2_STIG_SCAP_1-3_Benchmark-enhancedV8.xml.txt)

vanderpol:
With some updates to SCC's writing of OVAL results I'm now able to validate OVAL6 results as well:

```
14:36:55 - XML schema validation successful for:DESKTOP-7ODDPV4_SCC-5.10.1_Beta1_2024-12-10_143555_OVAL-CPE-Results_Microsoft_Windows_11_STIG-2.2.8.xml
14:36:57 - XML schema validation successful for:DESKTOP-7ODDPV4_SCC-5.10.1_Beta1_2024-12-10_143555_OVAL-Results_Microsoft_Windows_11_STIG-2.2.8.xml
14:36:57 - XML schema validation successful for:DESKTOP-7ODDPV4_SCC-5.10.1_Beta1_2024-12-10_143555_OVAL-Variables_Microsoft_Windows_11_STIG-2.2.8.xml

balleman-ctr:
We have a tool that uses JAXB for the OVAL schema.  I have updated the tool to use this branch of the OVAL schema, and with expected namespace changes, was able to build successfully.

vanderpol:
Thanks @balleman-ctr, sounds like we may be good enough for a merge of this PR.

PR #215, opened 2024-12-11
Replace the existing unused ESX schema with the CIS ESX extension

vanderpol:
I've updated the CIS ESX schema to comply with new OVAL 6.0 namespace, license, dates, etc...


vanderpol:
Files are schema valid and based on extensions used by CIS and Arctic Wolf

PR #216, opened 2024-12-11
Add kubernetes schema based on CIS OVAL extension

vanderpol:
Added schema based on CIS extension, updated to OVAL 6.0 namespaces, dates, and license


vanderpol:
Proceeding with merging this into develop.  The files are schema valid, and based on implementations from CIS and Arctic Wolf

PR #217, opened 2024-12-13
Updated documentation based on OVAL 6.0 schemas and updated examples

vanderpol:
Updated rst files, removing obsolete files, and showing an example of new encapsulated format.  Added new kubernetes


vanderpol:
Merging documentation change to develop so I can test it via readthedocs

Issue #181, opened 2024-10-22
Update structure of OVAL definitions to support encapsulated defintions and existing 5.x format


vanderpol:
**Abstract**
Update OVAL 6.0 to support a new proposed format of encapsulated definitions.  This is intended to increase adoption of OVAL and make content easier to maintain and merge with other content streams.  

The discussion can be viewed here:  https://github.com/orgs/OVAL-Community/discussions/162

The vote approving this change can be seen at https://github.com/OVAL-Community/OVAL-Board/blob/main/board_votes/2024-11-07_OVAL_6.0_Definition_Structure_Vote.md

**Link to Proposal**
Once a proposal has been put into pull request form, add a link to the PR here. If and as alternate proposals or objections are added they should be linked here as well.

**Additional context**
[Add any other context or screenshots about the enhancement.](https://github.com/orgs/OVAL-Community/discussions/162)

johnulmer-oval:
Have modified oval-definitions-schema.xsd in the 5.12_release branch to support encapsulated_definition(s).  Please see the notes attached to that commit.  The following is a example of content employing an encapsulate_definition.  

`<?xml version="1.0" encoding="UTF-8"?>
<oval_definitions
  xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
  xmlns:oval-def="http://oval.mitre.org/XMLSchema/oval-definitions-5"  
  xmlns:ind-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent"  
  xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5"
  xmlns:unix-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix"
  xmlns:linux-def="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux"
  xsi:schemaLocation="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent
    https://raw.githubusercontent.com/OVALProject/Language/5.11.2/schemas/independent-definitions-schema.xsd
  	http://oval.mitre.org/XMLSchema/oval-definitions-5#unix  https://raw.githubusercontent.com/OVALProject/Language/5.11.2/schemas/unix-definitions-schema.xsd
[...]

PR #219, opened 2024-12-16
Update oval-definitions-schema.xsd to support encapsulated definitions along with 5.8 style defintitions

vanderpol:
Committing the addition of the new encapsulated_definition and  EncapsulatedDefinitionType to preserve work and allow review.

Remaining to be addressed.
1 - In the EncapsulatedDefinitionType, the 'xsd:unique name="UniqueAffectedFamily"' is commented out.  Need to figure out best way to enforce this requirement. 2 - Review that modified 'testKeyRef' checks are appropriate.

### DIFF
--- a/oval-schemas/oval-definitions-schema.xsd
+++ b/oval-schemas/oval-definitions-schema.xsd
@@ -119,28 +119,28 @@
             <xsd:annotation>
                 <xsd:documentation>Requires each test reference to refer to a valid test id.</xsd:documentation>
             </xsd:annotation>
-            <xsd:selector xpath=".//*"/>
+            <xsd:selector xpath="encapsulated_definition/* | ./*" />
             <xsd:field xpath="@test_ref"/>
         </xsd:keyref>
         <xsd:keyref name="objectKeyRef" refer="oval-def:objectKey">
             <xsd:annotation>
                 <xsd:documentation>Requires each object reference to refer to a valid object id.</xsd:documentation>
             </xsd:annotation>
-            <xsd:selector xpath=".//*"/>
+            <xsd:selector xpath="encapsulated_definition/* | ./*" />
             <xsd:field xpath="@object_ref"/>
         </xsd:keyref>
         <xsd:keyref name="stateKeyRef" refer="oval-def:stateKey">
             <xsd:annotation>
                 <xsd:documentation>Requires each state reference to refer to a valid state id.</xsd:documentation>
             </xsd:annotation>
-            <xsd:selector xpath=".//*"/>
+            <xsd:selector xpath="encapsulated_definition/* | ./*" />
             <xsd:field xpath="@state_ref"/>
         </xsd:keyref>
         <xsd:keyref name="variableKeyRef" refer="oval-def:variableKey">
             <xsd:annotation>
                 <xsd:documentation>Requires each variable reference to refer to a valid variable id.</xsd:documentation>
             </xsd:annotation>
-            <xsd:selector xpath=".//*"/>
+            <xsd:selector xpath="encapsulated_definition/* | ./*" />
             <xsd:field xpath="@var_ref"/>
         </xsd:keyref>
         <xsd:keyref name="object_referenceKeyRef" refer="oval-def:objectKey">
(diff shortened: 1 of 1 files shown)


vanderpol:
@johnulmer-oval would like some feedback on his current attempts to support this.  Please review his commit notes/comments/questions.  There is some minor issue with metadata/UniqueAffectedFamily that he would particularly like some feedback on.  

Note that the branch used for development of this was created before we switched to the URN style namespaces, which may be easier for your debugging as well, less change.

Hoping everyone isn't already on vacation for the holidays.  This is our last 'big' item to be resolved before we finalize OVAL 6.0

vanderpol:
Sorry to beg for feedback, but this is pretty critical to OVAL 6.0, if you have a few minutes, we'd really like your input @solind, @A-Biggs, @maxullman @balleman-ctr

oval-schemas/oval-definitions-schema.xsd
```diff
@@ -239,6 +240,73 @@
...
+        <xsd:sequence>
+            <xsd:element ref="ds:Signature" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="metadata" type="oval-def:MetadataType">
+                <!-- xsd:unique name="UniqueAffectedFamily">
+                    <xsd:annotation>
+                        <xsd:documentation>Each affected element must have a unique family attribute value.</xsd:documentation>
+                    </xsd:annotation>
+                    <xsd:selector xpath="oval-def:affected"/>
+                    <xsd:field xpath="@family"/>
+                </xsd:unique -->
```
maxullman, line 276: What's the issue here for why this is commented out?  Unless I'm missing something, it seems no better or worse than the current validation for the old definition type.

vanderpol: @johnulmer-oval posted his answer on it back on the issue #181 (he's new to dealing with PR's and Github),  I'll re-post his comments here for easier reading.   Simple answer is that we just need to change them from being unique to elements, as they can no longer be validated as unique.

From @johnulmer-oval :
Have modified oval-definitions-schema.xsd in the '181-update-structure-of-oval-definitions-to-support-encapsulated-defintions...' branch to support encapsulated_definition(s).

Remaining to be addressed.
1 - In the EncapsulatedDefinitionType, the 'xsd:unique name="UniqueAffectedFamily"' is commented out. Need to figure out best way to enforce this requirement.
2 - Review that the modified 'testKeyRef' checks are appropriate.

An example of content employing an encapsulate_definition is attached.

Review and comment is requested.
[simplest_oval6_test.txt](https://github.com/user-attachments/files/18131615/simplest_oval6_test.txt)

As OVAL processing is unchanged from an OVAL 5.11 'definition' to an 'encapsulated_definition', OVAL results are typical.

@[johnulmer-oval](https://github.com/johnulmer-oval)
Member
johnulmer-oval commented [4 hours ago](https://github.com/OVAL-Community/OVAL/issues/181#issuecomment-2549592737)
RE: the 1st 'remaining' issue mentioned in my previous - the 'xsd:unique name="UniqueAffectedFamily"' in the EncapsulatedDefinitionType

When making the EncapsualteDefinitionType, I used the DefinitionType as a template. That brought along the metadata chunk and with it the xsd:unique for the UniqueAffectedFamily. This 'xsd:unique' cannot exist twice with the same element name. So, I'm trying to understand the intended use of the 'family'. After surfing as much documentation as I can find and some content streams, it appears to be mainly a reporting label. If this is the case, I think we can change the xsd:unique to a simple element and the non-uniqueness is eliminated. But, 'would appreciate any input others might have regarding the actual in practice use of the 'family'.

johnulmer-oval:
Re: the xsd:unique check on the family in the metadata, I moved the twin conflicting 'xsd:unique' checks from the 'definition' and 'encapsulated_definition' areas down into the 'metadataType'.

oval-schemas/oval-definitions-schema.xsd
maxullman: For all the content we've seen all the definitions in a given file have the same family.  I can only recall one time seeing a file with both family=unix definitions and some family=windows definitions in the same file, and even in that file each definition only had one family.  It sounds like @johnulmer-oval found a different solution but worst case I don't think it's exceptionally important the schema enforce anything for this attribute.

maxullman (approved):

PR #220, opened 2024-12-18
Added clarification documentation to windows ntuser test for improved consistency across OVAL interpreters

vanderpol:
Please review the documentation additions to the Windows 'ntuser' object and collected item.

We could also decide to add another behavior of something like 'only_include_interactive_users' with the default being 'true'


oval-schemas/windows-definitions-schema.xsd
```diff
@@ -1485,7 +1485,17 @@
       <xsd:element name="ntuser_object" substitutionGroup="oval-def:object">
             <xsd:annotation>
                   <xsd:documentation>The ntuser_object element is used to specify which metadata should be collected from a Windows ntuser.dat file.  Each object extends the standard ObjectType as defined in the oval-definitions-schema and one should refer to the ObjectType description for more information. The common set element allows complex objects to be created using filters and set logic. Again, please refer to the description of the set element in the oval-definitions-schema.</xsd:documentation>
-                  <xsd:appinfo>
+                  <xsd:documentation>To ensure consistent results across OVAL interpreters should comply with the following:</xsd:documentation>
+				  <xsd:documentation>1.  Only include ntuser.dat files from 'human' accounts that have interactively logged into the system, with the exception of the 'Default' user if the 'include_default' behavior is true.</xsd:documentation>
+				  <xsd:documentation>2.  Create a 'does not exist' item for each ntuser.dat file that does contain the required registry key.</xsd:documentation>
```
solind, line 1490: Should say "does not contain", I think.

solind: Although... you do know this requirement cannot impact test results, and therefore, should be more of a strong suggestion than a requirement.

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -1485,7 +1485,17 @@
...
+				  <xsd:documentation></xsd:documentation>
+				  <xsd:documentation>For requirement #1, OVAL interpreters are at their discretion to determine which ntuser.dat files have been logged into interactively, but the following is one method to determine this information.</xsd:documentation>
+				  <xsd:documentation>a.  Obtain a list of user profiles from the registry key:  'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList'</xsd:documentation>
```
solind, line 1493: Couldn't you just check that the user has logged on at some point?  Using whatever method is used to determine last_logon for the various user tests?  Might be simpler, since every interpreter already has to be able to collect that.

oval-schemas/windows-system-characteristics-schema.xsd
```diff
@@ -797,7 +797,10 @@
      <!-- =============================================================================== -->
      <xsd:element name="ntuser_item" substitutionGroup="oval-sc:item">
           <xsd:annotation>
-               <xsd:documentation>The windows ntuser_item specifies information that can be collected from a particular ntuser.dat file.</xsd:documentation>
+               <xsd:documentation>To ensure consistent results across OVAL interpreters should comply with the following:</xsd:documentation>
+			  <xsd:documentation>* Create a 'does not exist' item for each ntuser.dat file that does contain the required registry key.</xsd:documentation>
+			  <xsd:documentation>This requirement goes against normal item creation in OVAL, but is very useful in reporting to the end user which ntuser.dat files failed registry key existence checks when tests mandate 'all_exist'</xsd:documentation>
```
solind, line 802: I'd say it should be documented as "strongly suggested" rather than "required", since you can't write tests against items that don't exist (except that you can, of course, test that a particular item doesn't exist).

solind (changes requested):

oval-schemas/windows-system-characteristics-schema.xsd
vanderpol: The word 'required' is only related to a test looking for a required key, the overall update is just a strong recommendation with "should comply with'

To ensure consistent results across OVAL interpreters should comply with the following:

oval-schemas/windows-definitions-schema.xsd
vanderpol: Agreed on "does not contain", good catch.   As the overall documentation is a suggestion, I think we are OK?

To ensure consistent results across OVAL interpreters **should** comply with the following:

oval-schemas/windows-definitions-schema.xsd
solind: Yeah, I guess I do remember that you had to query the domain controller for domain accounts.  But, doesn't that give you the last time the user logged onto the domain?  We want the last time the user logged into the machine -- which is what you're capturing using a trick.  I wonder if that would capture logins using WS-Man/WinRM.  I don't have access to the method we used to capture the last_logon data, I'd need @maxullman or @A-Biggs to chime in.

vanderpol:
Thanks for your quick review @solind!  Please check to see if my fixes (and comments above regarding last_login) resolve your concerns, or if more changes are needed.

oval-schemas/windows-definitions-schema.xsd
vanderpol: If we could accurately determine the last_login relative to the machine being logged in, that would be a big win, and IMHO would warrant adding a 'last_login' attribute to the ntuser_item, as currently we just have last_modified, which can be inaccurate, as Windows updates sometimes arbitrarily update all ntuser.dat files on the system, making the file modified date worthless.

solind (commented):
I'd still like someone with access to an interpreter implementation to check on that "has ever logged in" thought.

Honestly, I think any domain user with an ntuser.dat file on a machine must have logged on at some point.

oval-schemas/windows-system-characteristics-schema.xsd
```diff
@@ -798,9 +798,9 @@
...
+			  <xsd:documentation>* Create a 'does not exist' item for each ntuser.dat file that does not contain the specified registry key.</xsd:documentation>
 			  <xsd:documentation>This requirement goes against normal item creation in OVAL, but is very useful in reporting to the end user which ntuser.dat files failed registry key existence checks when tests mandate 'all_exist'</xsd:documentation>
-			  <xsd:documentation>Example:  If a computer contains 3 user profiles of human users, and one user has the registry key/name/value required, and is configured correctly, but the other 2 users do not have the required registry key, creating 'does not exist' items, informs the end user, which ntuser.dat files, (which corresponds to which specific user(s)) are not compliant.</xsd:documentation>
+			  <xsd:documentation>Example:  If a computer contains 3 user profiles of human users, and one user has the registry key/name/value specified, and is configured correctly, but the other 2 users do not have the required registry key, creating 'does not exist' items, informs the end user, which ntuser.dat files, (which corresponds to which specific user(s)) are not compliant.</xsd:documentation>
```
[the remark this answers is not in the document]
vanderpol: good catch again.  Will fix

vanderpol:
BTW, @solind  you got me thinking... we might be able to capture the lastModified of the \Software\Microsoft\Windows\CurrentVersion\Explorer key, as the LoginCount data is incremented with each login.  Below is a sample size of 1, but seems to be accurate.

Key Name:          HKEY_USERS\S-1-5-21-9063620-3197651053-1651582182-1000\Software\Microsoft\Windows\CurrentVersion\Explorer
Class Name:        <NO CLASS>
Last Write Time:   12/18/2024 - 8:40 AM

This data could be useful as a new element on the collected item, but checking to see if the LoginCount or UserSignedIn is still required IMHO.  We had an end user that had some management software that was used to arbitrary command remotely and non-interactively with domain profiles.  Windows created partial/stub profiles for each user that ran these commands, but as they never logged in interactively, things like Group Policy are never applied, and would cause all of their systems to fail all Office tests. The "Explorer" key might still be updated, but they keys showing a login did not exist.

oval-schemas/windows-definitions-schema.xsd
vanderpol: I think I've resolved this feedback item?

solind:
I know we've had similar issues with inactive user profiles causing what users considered to be false-positives.  But, sometimes they also yield actual positives!  There's a fair amount of knowledge/experience encapsulated in the implementation, IIRC, but I haven't had access to that code in a long time.

vanderpol:
> I know we've had similar issues with inactive user profiles causing what users considered to be false-positives. But, sometimes they also yield actual positives! There's a fair amount of knowledge/experience encapsulated in the implementation, IIRC, but I haven't had access to that code in a long time.

I'm definitely open to a shared knowledge base on this, and allow tools to be more consistent.  I know that excluding users based on certain scenarios comes with risk, but when we had users complaining about partial/stub/orphaned profiles causing false positives, that their only fix was to delete the offending profiles, my thought was that if a user hasn't ever logged into Windows Explorer, and has no chance of getting GPO's applied, those likely are safe to exclude.  I'm just sharing what we implemented as a suggestion in hopes that content could run the same across tools.   If someone else has a better/more accurate way, I'm open to all ideas.

oval-schemas/windows-system-characteristics-schema.xsd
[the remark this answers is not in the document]
vanderpol: and back to your "any domain user with an ntuser.dat file on a machine must have logged on at some point' statement.  6 months ago I would have agreed, but had end users share ntuser.dat files that proved otherwise.  I diffed their stub ntuser.dat file which was created by some 3 party application that spawns commands and returns data back to a remote computer, and found that there can be ntuser.dat files that are residue from that process.  It might also be possible create stub ntuser.dat files with remote powershell commands based on options provided as well.

In my comparison of a 'normal' ntuser.dat file and the stub file from an end user, is where I found the differences with the Explorer and "UserSignedIn" value.   The value doesn't exist at all with this scenario, and is the basis of this documentation update based on our implementation.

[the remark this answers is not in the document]
solind: Interesting.  I bet these are artifacts from services that are configured to run using domain accounts.

I know in the past I've seen these as residue from bad image-based installations with bad sysprep, but it's been a long time, and I always considered that a real hygiene problem that customers had and should fix.

PR #226, opened 2025-01-14
Added AIX interim_fix, os_level tests and Solaris package, package511 and patch54 tests back to OVAL 6

vanderpol:
Due to documented usage from SecPod


Pooja-SecPod:
I have checked these are the probes required. I am not sure how do I mark as reviewed.

vanderpol:
> I have checked these are the probes required. I am not sure how do I mark as reviewed.

I'll take that as a confirmation of review, and merge this PR

PR #231, opened 2025-01-21
Updated shellcommand documentation to make it clear that the 'command' is not a script

vanderpol:
The 'command' of the test is intended to be a literal executable/binary, or a one-liner where data from one or more executables is manipulated and the intended text is printed to standard out.


johnulmer-oval (approved):
New language makes sense.  The capture of exit codes/status from multi-line scripts is problematic as it is obscures the origin of the code received.

PR #243, opened 2025-02-27
Resolved #242 by adding full_name, comment and password_age_days to user_sid55 state and item

### LINKED ISSUE #242: Add  'full_name', 'comment', and 'password_age_days' attributes to the Windows user_sid55 state and item
These attributes were added to the Windows 'user' state and item in 5.11.2, however that test was also deprecated in 5.11.2, replaced by the user_sid55 test, but the user_sid55 test was not updated to include these attributes.

In order to maintain the same capability as OVAL 5.11.2, these attributes need to be added to the user_sid55 test.

vanderpol:
In order to maintain functionality from OVAL 5.11.2, these elements are needed, as we dropped the windows user test

### DIFF
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -2692,37 +2692,52 @@
             <xsd:complexContent>
                 <xsd:extension base="oval-def:StateType">
                     <xsd:sequence>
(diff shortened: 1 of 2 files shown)


solind (approved):
Looks fine!

PR #257, opened 2025-04-10
Added days_since_last_logon attribute to resolve #221

### LINKED ISSUE #221: Add 'days_since_last_logon' to the Windows ntuser item and state
**Abstract**
When the ntuser test was proposed, there was not a method that seemed possible to determine when a give user (especially a domain user) last logged into the system.  In order to provide some data hopefully related to this a "date_modified" value was added, but this is just the last time the ntuser.data file was modified.  Usually this is the same time as the user logged in, but other Windows processes (such as WSUS) can modify all ntuser.dat files, making the data near meaningless.

As part of the PR review for #220, @solind  mentioned there likely exists a method to determine when a domain user last logged into a target computer.  This ticket is pending any feedback from @A-Biggs  or @maxullman  who may have some additional insight.

@vanderpol noticed that inside each user's ntuser.dat file, the HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer "logonCount" registry value is incremented on each login, and the last write time for the Explorer key is updated, but once again this could be updated by other updates to the Explorer key, and may not be 100% reliable.

Example:

**Link to Proposal**
Once a proposal has been put into pull request form, add a link to the PR here. If and as alternate proposals or objections are added they should be linked here as well.

**Additional context**
Add any other context or screenshots about the enhancement.

vanderpol:
Added element per discussions with @solind

### DIFF
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -4066,7 +4066,7 @@
                                     </xsd:element>
 								  <xsd:element name="days_since_last_logon" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
 									   <xsd:annotation>
-											<xsd:documentation>The last_logon data, converted to days and then rounded to an integer.  If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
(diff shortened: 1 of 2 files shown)


solind (commented):
Is "rounded" too ambiguous?  I think if it's been < 24 hrs since the last login, perhaps this should be 0, less than 48 hrs than 1, etc.  Kind of like the proverbial "days since last accident" sign on a jobsite.

solind:
Works for me.

solind (approved):

Issue #251, opened 2025-03-18
Update version and dates for OVAL 5.12.1 release


vanderpol:
Update versions in all files and dates to be 

5.12.1
05/22/2025

PR #265, opened 2025-05-20
Updating version and date for #251 pending oval board approval on 05/22/2023

vanderpol:
Updated version to 5.12.1 and date to 05/23/2023, pending oval board approval on 05/22/2023 board meeting.


solind (approved):
Sometimes when I see these kinds of commits I think, we should templatize the XML and keep the version and date in a properties file.  But it happens so infrequently it never seems worthwhile...

vanderpol:
> Sometimes when I see these kinds of commits I think, we should templatize the XML and keep the version and date in a properties file. But it happens so infrequently it never seems worthwhile...

I would agree, and maybe we should revisit this in OVAL 6?

PR #274, opened 2025-06-02
Added '' to allow for variable usage for Windows wuaupdateserver

vanderpol:
This was missed in our update to 5.12.1


solind (approved):
Oh yeah, all enumerations should allow for an empty string value to accommodate variable references.

PR #290, opened 2025-09-26
Resolved #288 by fixing maxOccurs from 1 to unbounded

### LINKED ISSUE #288: Fix OVAL IOS XE section state to update config_line MaxOccurs from 1 to unbounded
The IOS XE section test has a bug, as documentation states that the config_line should be 'one config line of the section at a time' but then limits it to one, which is contradictory.

vanderpol:
Fixing issue with the state, see related issue #288

### DIFF
--- a/oval-schemas/iosxe-definitions-schema.xsd
+++ b/oval-schemas/iosxe-definitions-schema.xsd
@@ -679,7 +679,7 @@
                                 <xsd:documentation>The value returned with all config lines of the section.</xsd:documentation>
                             </xsd:annotation>
                         </xsd:element>
-                        <xsd:element name="config_line" type="oval-def:EntityStateStringType" minOccurs="0" maxOccurs="1">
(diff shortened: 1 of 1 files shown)


solind (commented):
Oh wait!  I should not have approved this.  A state actually should only specify a single `config_line` entity!

vanderpol:
Good catch, I think we made the same mistake on the shell command state.  This update should.habw been applied to the section item, as the config line should allow for 0-N lines 

Should I fix the shell command state as well?

solind:
Yeah, absolutely, if there's the same issue.  The state can have unbounded items, but the state never can, because there's then no sensible way to compare the state to an item. (Of course, you can have multiple states in a test, which is how you're supposed to handle that scenario).

vanderpol:
@solind  Closing issue, without merging any changes, and deleting branch, as no change is required.  See #292 for related fix to shellcommand state.

Issue #308, opened 2026-04-22
Update Windows ntuser test to mandate creation of collected item for a given ntuser.dat file even if the registry key/name does not exist


vanderpol:
Currently, some of the STIG rules stipulate that a for a user to be compliant, a certain registry value needs to exist and be set properly.  However, the ntuser_test doesn’t have a way to enforce that a given registry key must exist for each user on the system.  If one compliant user is found, an ntuser_item is collected for that user, and non-compliant users are skipped (so long as the key of interest is not present – which is often the default), and the rule is given a pass. 

In NIWC's SCC application, they force the creation of a 'does not exist' item, which is allowed in OVAL and allows content to then report correctly, but is not being implemented this way in other tools, leading to inconsistently results.

PR #310, opened 2026-05-08
Updated ntuser documentation based on implementation docs and clarified when an item should be created and added new behavior per #307 and #308

### LINKED ISSUE #307: Update Windows ntuser documentation based on original specification/implemention documents
**Abstract**
Given the complexity of this check, in order to ensure consistent results across implementations, the details of how to obtain the user profiles should be included.  This will also help adoption for future vendors.

vanderpol:
Updated docs to clear document how to go about determining which ntuser.dat files should be included, and when to create a ntuser_item

### DIFF
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -4064,7 +4064,9 @@
                                     <xsd:element name="logged_on" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The logged_on element describes if the user account is currently logged on to the computer.</xsd:documentation>
-												<xsd:documentation>This can be determined by comparing the SID’s against those populated in HKEY_USERS.</xsd:documentation>
+												<xsd:documentation>This can be determined by comparing the SIDs collected from the ProfileList against those populated in HKEY_USERS\&lt;SID&gt;</xsd:documentation>
+												<xsd:documentation>HKEY_USERS: Contains all the actively loaded user profiles on the computer.  https://learn.microsoft.com/en-us/troubleshoot/windows-server/performance/windows-registry-advanced-users</xsd:documentation>
+												<xsd:documentation>This data can also be obtained by other various Windows API's such as a combination of win32_logonsession and win32_loggedonuser, but the specifics are beyond the scope of OVAL documentation.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
 								  <xsd:element name="days_since_last_logon" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -3931,6 +3931,8 @@
                   <xsd:documentation>The ntuser test is used to check metadata associated with Windows ntuser.dat files. It extends the standard TestType as defined in the oval-definitions-schema and
                         one should refer to the TestType description for more information. The required object element references a ntuser_object and the optional state element specifies the ntuser
                         data to check.</xsd:documentation>
+						
+						<xsd:documentation>To ensure consistent results across OVAL interpreters, application developers should refer to documentation in the ntuser_item when implenting this test.</xsd:documentation>
                   <xsd:appinfo>
                         <oval:element_mapping>
                               <oval:test>ntuser_test</oval:test>
@@ -4035,11 +4037,13 @@
                                     <xsd:element name="key" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes a registry key normally found in the HKCU hive to be tested.</xsd:documentation>
+												<xsd:documentation>Note:  The key not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="name" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the name of a value of a registry key.</xsd:documentation>
+												<xsd:documentation>Note:  The name not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="sid" type="oval-def:EntityStateStringType" minOccurs="0">
@@ -4062,16 +4066,18 @@
                                     <xsd:element name="logged_on" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The logged_on element describes if the user account is currently logged on to the computer.</xsd:documentation>
+												<xsd:documentation>This can be determined by comparing the SID’s against those populated in HKEY_USERS</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
 								  <xsd:element name="days_since_last_logon" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
 									   <xsd:annotation>
-											<xsd:documentation>The last_logon data, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
+											<xsd:documentation>The last_logon data which can be obtained from the LocalProfileLoadTimeHigh and LocalProfileLoadTimeLow values, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
 									   </xsd:annotation>
 								  </xsd:element>
                                     <xsd:element name="enabled" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The enabled element describes if the user account is enabled or disabled.</xsd:documentation>
+												<xsd:documentation>Note: For domain users, if a domain controller is not available, this will not return data.  If using this data for a filter to include enabled accounts, it’s recommended to exclude accounts that are have been determined to be disabled, vs including ones that are enabled, as the later may filter out accounts for which the domain controller could not return data.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="date_modified" type="oval-def:EntityStateIntType" minOccurs="0">
@@ -4088,6 +4094,7 @@
                                     <xsd:element name="filepath" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the filepath of the ntuser.dat file.</xsd:documentation>
+												<xsd:documentation>The existance of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="last_write_time" type="oval-def:EntityStateIntType" minOccurs="0">
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -4037,13 +4037,11 @@
                                     <xsd:element name="key" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes a registry key normally found in the HKCU hive to be tested.</xsd:documentation>
-												<xsd:documentation>Note:  The key not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="name" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the name of a value of a registry key.</xsd:documentation>
-												<xsd:documentation>Note:  The name not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="sid" type="oval-def:EntityStateStringType" minOccurs="0">
@@ -4066,18 +4064,19 @@
                                     <xsd:element name="logged_on" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The logged_on element describes if the user account is currently logged on to the computer.</xsd:documentation>
-												<xsd:documentation>This can be determined by comparing the SID’s against those populated in HKEY_USERS</xsd:documentation>
+												<xsd:documentation>This can be determined by comparing the SID’s against those populated in HKEY_USERS.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
 								  <xsd:element name="days_since_last_logon" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
 									   <xsd:annotation>
-											<xsd:documentation>The last_logon data which can be obtained from the LocalProfileLoadTimeHigh and LocalProfileLoadTimeLow values, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
+											<xsd:documentation>The last_logon data which can be obtained from the LocalProfileLoadTimeHigh and LocalProfileLoadTimeLow registry values from HKLM\Software\Microsoft\Windows NT\CurrentVersion\ProfileList\&lt;SID&gt;, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
+											<xsd:documentation>For more information, refer to https://learn.microsoft.com/en-us/troubleshoot/windows-server/support-tools/scripts-to-retrieve-profile-age</xsd:documentation>
 									   </xsd:annotation>
 								  </xsd:element>
                                     <xsd:element name="enabled" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The enabled element describes if the user account is enabled or disabled.</xsd:documentation>
-												<xsd:documentation>Note: For domain users, if a domain controller is not available, this will not return data.  If using this data for a filter to include enabled accounts, it’s recommended to exclude accounts that are have been determined to be disabled, vs including ones that are enabled, as the later may filter out accounts for which the domain controller could not return data.</xsd:documentation>
+												<xsd:documentation>Note: For domain users, if a domain controller is not available, this will not return data, and should be reported with a status of 'not collected'.  If using this data for a filter to include enabled accounts, it’s recommended to exclude accounts that are have been determined to be disabled, vs including ones that are enabled, as the later may filter out accounts for which the domain controller could not return data.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="date_modified" type="oval-def:EntityStateIntType" minOccurs="0">
@@ -4094,7 +4093,6 @@
                                     <xsd:element name="filepath" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the filepath of the ntuser.dat file.</xsd:documentation>
-												<xsd:documentation>The existance of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="last_write_time" type="oval-def:EntityStateIntType" minOccurs="0">
@@ -4191,6 +4189,20 @@
                         </xsd:restriction>
                   </xsd:simpleType>
             </xsd:attribute>
+            <xsd:attribute name="item_creation" use="optional" default="key_and_name_exist">
+                  <xsd:annotation>
+                        <xsd:documentation>For 'key_and_name_exist', items are only created when an ntuser.dat file includes the key and name provided in the ntuser object.</xsd:documentation>
+                        <xsd:documentation>For 'every_ntuser', items are created for each relavent ntuser.dat found on the system. This option will prevent false negatives in instances where each ntuser.dat file must contain the required key/name/value in order to pass, but the file is lacking the key/name required to normally satisfy the creation of an ntuser item.</xsd:documentation>
+                        <xsd:documentation></xsd:documentation>
+                  </xsd:annotation>
+                  <xsd:simpleType>
+                        <xsd:restriction base="xsd:string">
+                              <xsd:enumeration value="key_and_name_exist"/>
+                              <xsd:enumeration value="every_ntuser"/>
+                        </xsd:restriction>
+                  </xsd:simpleType>
+            </xsd:attribute>			
+
       </xsd:complexType>
       <!-- =============================================================================== -->
       <!-- ===========================  PASSWORD POLICY TEST  ============================ -->
--- a/oval-schemas/windows-system-characteristics-schema.xsd
+++ b/oval-schemas/windows-system-characteristics-schema.xsd
@@ -1680,7 +1680,8 @@
                               <xsd:element name="logged_on" type="oval-sc:EntityItemBoolType" minOccurs="0" maxOccurs="1">
                                    <xsd:annotation>
                                         <xsd:documentation>The logged_on element describes if the user account is currently logged on to the computer.</xsd:documentation>
-										<xsd:documentation>This can be determined by comparing the SIDs collected from the ProfileList against those populated in HKEY_USERS</xsd:documentation>
+										<xsd:documentation>This can be determined by comparing the SIDs collected from the ProfileList against those populated in HKEY_USERS\&lt;SID&gt;</xsd:documentation>
+										<xsd:documentation>HKEY_USERS: Contains all the actively loaded user profiles on the computer.  https://learn.microsoft.com/en-us/troubleshoot/windows-server/performance/windows-registry-advanced-users</xsd:documentation>
                                    </xsd:annotation>
                               </xsd:element>
                               <xsd:element name="days_since_last_logon" type="oval-sc:EntityItemIntType" minOccurs="0" maxOccurs="1">
--- a/oval-schemas/windows-system-characteristics-schema.xsd
+++ b/oval-schemas/windows-system-characteristics-schema.xsd
@@ -1711,7 +1711,6 @@
                               <xsd:element name="filepath" type="oval-sc:EntityItemStringType" minOccurs="0" maxOccurs="1">
                                    <xsd:annotation>
                                         <xsd:documentation>This element describes the filepath of the ntuser.dat file.</xsd:documentation>
-										<xsd:documentation>The existance of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
                                    </xsd:annotation>
                               </xsd:element>
                               <xsd:element name="last_write_time" type="oval-sc:EntityItemIntType" minOccurs="0" maxOccurs="1">
--- a/oval-schemas/windows-system-characteristics-schema.xsd
+++ b/oval-schemas/windows-system-characteristics-schema.xsd
@@ -1688,13 +1688,7 @@
 										<xsd:documentation>The last_logon data which can be obtained from the LocalProfileLoadTimeHigh and LocalProfileLoadTimeLow registry values from HKLM\Software\Microsoft\Windows NT\CurrentVersion\ProfileList\&lt;SID&gt;, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
 										<xsd:documentation>For more information, refer to https://learn.microsoft.com/en-us/troubleshoot/windows-server/support-tools/scripts-to-retrieve-profile-age</xsd:documentation>
 								   </xsd:annotation>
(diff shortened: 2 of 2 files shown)


oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4035,11 +4037,13 @@
                                     <xsd:element name="key" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes a registry key normally found in the HKCU hive to be tested.</xsd:documentation>
+												<xsd:documentation>Note:  The key not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
                                           </xsd:annotation>
                                     </xsd:element>
                                     <xsd:element name="name" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the name of a value of a registry key.</xsd:documentation>
+												<xsd:documentation>Note:  The name not existing on the target does not impact the overall existence of the ntuser_item.</xsd:documentation>
```
solind: If there is an item with a corresponding name, the object will be said to exist.

Anyway, this is documentation for the state entity. If the state specifies a name that doesn't match any item, then no items will match that state. The effect on existence comes into play only if the state is used to filter an object or set.

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4062,16 +4066,18 @@
                                     <xsd:element name="logged_on" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The logged_on element describes if the user account is currently logged on to the computer.</xsd:documentation>
+												<xsd:documentation>This can be determined by comparing the SID’s against those populated in HKEY_USERS</xsd:documentation>
```
solind: The active user logon sessions can also be determined using the WMI classes Win32_LoggedOnUser and Win32_LogonSession. I'm not sure whether the presence of a hive under HKEY_USERS is completely reliable for the purpose of whether the user should actually be considered "logged on".

I recall we considered interactive and remote interactive session types to be logged on.

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4062,16 +4066,18 @@
...
                                           </xsd:annotation>
                                     </xsd:element>
 								  <xsd:element name="days_since_last_logon" type="oval-def:EntityStateIntType" minOccurs="0" maxOccurs="1">
 									   <xsd:annotation>
-											<xsd:documentation>The last_logon data, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
+											<xsd:documentation>The last_logon data which can be obtained from the LocalProfileLoadTimeHigh and LocalProfileLoadTimeLow values, converted to days and then rounded down to the nearest integer (floor function). If the account is determined to be currently logged in, this date should be reported as 0.</xsd:documentation>
```
solind: Where are these obtained from?  Registry values somewhere under HKLM?

Data can also be obtained from the WMI table Win32_NetworkLoginProfile.

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4062,16 +4066,18 @@
...
 									   </xsd:annotation>
 								  </xsd:element>
                                     <xsd:element name="enabled" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>The enabled element describes if the user account is enabled or disabled.</xsd:documentation>
+												<xsd:documentation>Note: For domain users, if a domain controller is not available, this will not return data.  If using this data for a filter to include enabled accounts, it’s recommended to exclude accounts that are have been determined to be disabled, vs including ones that are enabled, as the later may filter out accounts for which the domain controller could not return data.</xsd:documentation>
```
solind: I think we'd have to say that the element would be `status="not_collected"` (IIRC) if the domain server is unavailable, which (for good or ill) has documented evaluation effects.  Or, it could have an error status.

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4088,6 +4094,7 @@
                                     <xsd:element name="filepath" type="oval-def:EntityStateStringType" minOccurs="0">
                                           <xsd:annotation>
                                                 <xsd:documentation>This element describes the filepath of the ntuser.dat file.</xsd:documentation>
+												<xsd:documentation>The existance of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
```
solind: I hate to get into semantics, but I think you mean to be talking about the existence of the object, not the item.  The object exists if all its criteria are satisfied by one or more items discovered on the target machine.

oval-schemas/windows-system-characteristics-schema.xsd
```diff
@@ -1629,6 +1629,22 @@
...
+<xsd:documentation>		i.	HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\WindowsNT\CurrentVersion\ProfileList</xsd:documentation>
+<xsd:documentation>	 b.	Determine which user profiles are from ‘human’ users</xsd:documentation>
+<xsd:documentation>		i.	Exclude profiles for LocalService, Network service etc.. by excluding subkeys in the format of S-1-5-&lt;number*gt;</xsd:documentation>
+<xsd:documentation>		ii.	Exclude profiles for Per-Service SIDs by excluding subkeys in the format of S-1-5-80-&lt;number*gt;-&lt;number*gt;-&lt;number*gt;-&lt;number*gt;-&lt;number*gt;</xsd:documentation>
+<xsd:documentation>		iii. Include Local and Domain User SIDs by including subkeys match the format of S-1-5-21-&lt;number*gt;-&lt;number*gt;-&lt;number*gt;-&lt;number*gt;</xsd:documentation>
+<xsd:documentation>	 c.	Obtain ntuser filepath from the ProfileImagePath value of 'human' profiles</xsd:documentation>
+<xsd:documentation>2.	Creating ntuser items</xsd:documentation>
+<xsd:documentation>If the filepath obtained from the ProfileImagePath exists on the target system, create a ntuser_item with a status of 'exists'</xsd:documentation>
```
solind: This is basically saying that all NTUSER objects always exist, assuming there is at least one user profile on the target machine (there should always be at least one). I don't think that's right, but... it's been a long time.

oval-schemas/windows-system-characteristics-schema.xsd
```diff
@@ -1656,6 +1674,7 @@
                                         <xsd:documentation>The username entity holds a string that represents the name of a particular user. In Windows, user names are case-insensitive. As a
                                              result, it is recommended that the case-insensitive operations are used for this entity. In a domain environment, users should be identified in
                                              the form: "domain\user name". For local users use: "computer name\user name".</xsd:documentation>
+										<xsd:documentation>Note: When gathering the built-in Guest and build-in Administrator, they may not resolve and may need to have the ComputerName prepended to it.</xsd:documentation>
```
solind: The built-in guest and administrator accounts may actually be renamed. They are really only recognizable using their SIDs. I believe the username convention here should follow those used for other Windows user objects.

solind (changes requested):
I'm not quite sure about these documentation changes. In particular, those that infer something about the existence of ntuser_objects based on state entities.

oval-schemas/windows-definitions-schema.xsd
vanderpol: Thanks Dave, and yes this is likely a copy/paste issue from the item, I'll clear it up.  Any discussions on existence should only be in the item documentation.

oval-schemas/windows-definitions-schema.xsd
vanderpol: @solind good points, I'll make some updates.  Our rationale for uisng HKEY_USERS was two fold, if a user is logged in, their ntuser.dat file is locked, and the only way to get the data is to find it via HKEY_USERS\<SID>.    For 'interactive' and 'remote interactive', I assume that means 'local' and 'remote desktop'?   If not, please elaborate.  I may have to ponder if these keys are populated for remote ssh/winRm type sessions, as thay may not be.

oval-schemas/windows-definitions-schema.xsd
vanderpol: @solind I'll update the documentation to list either the currently logged in users ntuser.dat file or Win32_NetworkLoginProfile.

oval-schemas/windows-definitions-schema.xsd
vanderpol: Agreed, will update.

oval-schemas/windows-definitions-schema.xsd
vanderpol: I think this overlaps with your next comment, so I'll consolidate our discussions there.

oval-schemas/windows-system-characteristics-schema.xsd
vanderpol: @solind this is definitely the key point of discussion for all of these updates, and the primary point of concern about consistency across products.  Somehow we need to have a way to accurately perform checks for N ntuser.dat files.  Our team feels that one key documentation item that is lacking from OVAL, and is on our list of planned updates is to document when an item should be created.   Most are obvious and intuitive, but others are a bit more open for discussion/debate.

OVAL also doesn't handle N instances very well with existence.  The ntuser test and TextFileContent54 and XMLFileContents all suffer from this.  If you need to confirm that "ALL" instances are configured correctly, but the matching pattern (textfilecontent54 and xmlcontents) do not exist in a given file, then they are excluded from comparison, which works OK if there is only ONE file in scope, but if N files are in scope and 1 files has the required pattern, and is configured correctly and say 10 other file exist that should be configured do not have the requisite pattern, then the test passes, but should fail.   The same issue exist with the NTUSER test due to the N ntuser items issue.  If there are for simplicity's sake 2 users on the system, and one user has the required registry key/value and matches the state requirement and passes the test, but there is another user profile that does not have the required key or name, then in some OVAL interpreters, no item is created, and then the overall test passes, but it should fail because one of the nuser files is incorrectly configured.

From my view, we have 2 options:

1. Mandate the creation of 'does not exist' items if they ntuser filepath exists, but the required key/value does not exist.
2. Mandate the creation of an item with a status of 'exists', documenting that the key to existence is the 'filepath'.

What we did historically was to create items with a status of 'does not exist', which made the results accurate, but goes against the norms of OVAL which says to only create 'does not exist' items for debugging purposes.   You can see an example in the attached.

I'm open to any other suggestions you may have, but some changes is needed, as some of our end users have reported different results from our SCC tool to Trellix Policy Auditor, as Trellix does not create items ('exist' or 'does not exist') and the overall results differ when at least one ntuser.dat file is not configured at all via GPO's etc..  

The obvious side effect of mandating an ntuser_item being 'exist' is that existence only tests without a state won't be possible, or useful, but given the nature of this test, I don't see much of any use in that scenario, and the existence testing can then be moved down to the element in a state, which is what we do for the commandshell test, which by design always and only creates 1 item.
[ntuser_existence_example.xml](https://github.com/user-attachments/files/27592729/ntuser_existence_example.xml)

oval-schemas/windows-system-characteristics-schema.xsd
vanderpol: Agreed, and my proposed wording was poorly chosen, I was trying to say that Local Admin or Guest accounts (renamed or not) do not resolve to Hostname\\Account they just resolve to <Account> and the local hostname scope needs to be prepended to it.  I'll check other documentation to see how it's handled in other places.

oval-schemas/windows-definitions-schema.xsd
solind: Here is a link to the documentation for that WMI class: https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-logonsession

Interactive should include WinRM sessions, whereas remote interactive means a Windows terminal service session.  Remember Windows terminal services?  Is that still even a thing?  Somewhere, probably...

oval-schemas/windows-system-characteristics-schema.xsd
solind: Well, a non-existing item can never have an impact on the result. That much I know for sure.

When we implemented various CIS benchmarks in pure OVAL we used to jump through a bunch of hoops in such instances. We'd create an object to count the number of instances of the file (for example) and then compare it to the unique set of filenames with content matching the pattern in the context of a complicated OVAL definition.  It wasn't easy... but we made it fairly easy to do using our Slang language/tool (IDK if that still really even exists now, though).  I prefer that approach over mandating the creation of so-called partial matches. It's trivially easy to create an ntuser_obect that's guaranteed to exist for every user:

```
<ntuser_object>
  <key>/</key>
  <value/>
</ntuser_object>
```

solind: One more thing...

> The obvious side effect of mandating an ntuser_item being 'exist' is that existence only tests without a state won't be possible, or useful, but given the nature of this test

Actually the result will be that it would lead to errors, as states would be compared to items with field entities that were not collected or don't exist, when they were not necessarily designed with that possibility in mind.

vanderpol: @solind I'm not following how your recommendations solve the issue at hand of verifying that all user profiles need to have a required key/name/value to pass.  Can you provide an example? Maybe using Joval or SCC and attach some results?

vanderpol: So as you have pointed out, our existing method for creating a "does not exist' item and then actually using it to report pass fail based on the state is incorrect.  At a minimum, if we were to proceed with the 'does not exist' item creation, the failure would then be based on the test check_existence of "all_exist"  (which happens to be how most of the ntuser_tests are defined by DISA currently).

I am going to experiment with updating our code to create an existing item, based on the filepath existing.  The spec allows for 0-1 on Key/Name and 0-N on Value, so it's perfectly valid in the schema to not populate those elements, and other test types exist where not all elements may be populated for a given item.

I feel strongly that some change needs to be implemented, otherwise false negatives will (and are currently) occuring with most vendor implementations.  If you have a different/better method I'm open to all suggestions, but I'm also OK with documenting a slightly abnormal use case for thist test to ensure accuracy and consistency across products.

solind: Hi @vanderpol, that's something of a tall order without any authoring tools, but let me pseudo-code it and you can tell me if that helps.

Imagine I have two ntuser_objects:

```
<ntuser_object id="oval:ntuser:obj:1" comment="all user accounts">
  <key>/</key>
  <name/>
</ntuser_object>
<ntuser_object id="oval:ntuser:obj:2" comment="The foo user configuration">
  <key>/foo</key>
  <name>bar</name>
</ntuser_object>
```

And, I will have a state that checks the value of the second object:

```
<ntuser_state id="oval:ntuser:ste:2" comment="User configuration state /foo@bar=baz">
  <key>/foo</key>
  <name>bar</name>
  <value>baz</value>
</ntuser_state>
```

A simple test will nominally check the values for all configured users:

```
<ntuser_test id="oval:ntuser:tst:2" comment="user accounts are foo-configured">
  <object>oval:ntuser:obj:2</object>
  <state>oval:ntuser:ste:2</state>
</ntuser_test>
```

As you've pointed out, that test alone is insufficient, as users who don't have a foo configuration will not have generated items, and therefore, will be missed by the test.  To compensate for this, we need to employ variables, as follows (here I am counting unique SIDs, just to be safe):

```
<local_variable id="oval:ntuser:var:1" comment="total number of users">
  <count>
    <unique>
      <object_component object_ref="oval:ntuser:obj:1" item_field="sid" />
    </unique>
  </count>
</local_variable>
<local_variable id="oval:ntuser:var:2" comment="number of foo-configured users">
  <count>
    <unique>
      <object_component object_ref="oval:ntuser:obj:2" item_field="sid" />
    </unique>
  </count>
</local_variable>
```

Now, to test that the number of configured users matches the number of total users, we need a variable test:

```
<variable_object id="oval:ntuser:obj:3" comment="Variable containing the total number of users">
  <var_ref>oval:ntuser:var:1</var_ref>
</variable_object>
...
<variable_state id="oval:ntuser:ste:3" comment="Variable containing the number of configured users">
  <var_ref>oval:ntuser:var:2</var_ref>
</variable_state>
...
<variable_test id="oval:ntuser:tst:3" comment="The number of configured users equals the number of total users">
  <object>oval:ntuser:obj:3</object>
  <state>oval:ntuser:ste:3</state>
</variable_test>
```

Now, all you need is to write a definition that ANDs together the two tests, and voila.

Does that make sense?

vanderpol: @solind Thanks for the clarification, and I agree that we could get to the point of knowing if the system passes or fails, but the results will be meaningless to the end users.  Imagine a computer that has 20, 100 or 1000 ntuser.dat profiles on it, and one of those profiles does not have the needed key/name present.   The test then fails because say 99 is less than 100.   Now the end user gets the fun of trying to figure out which user profile that is, which they cannot do without a tool that can load all of the user profiles.  (such as the tool which just scanned the target)

The results from the scan need to state which ntuser.dat file, and all of the related metadata (username, domain, hostname etc... ) so they can figure out how to make it compliant.  How to fix a domain user is different from a local user etc... and you need to know exactly which user to fix.

So while your method is techincally valid, it's not useful for the level of detail needed.   We need to have an item created and populated with as much metadata as possible.   The 'more right' method seems from my perspective to be an item that exists, based on the filepath of the ntuser.dat file, but creating a 'does not exist' item, and then having all tests be check_existence = "all_exist" would produce the same overall effect.

vanderpol: I'm adding @balleman-ctr  and @jspearman19 from DISA to see what their opinion is, as they are involved in SCAP content writing.  I would include Graham Burkholder, but he's not currently on the OVAL board.  We have been in discussions about the ntuser test and differences in results with SCC and Policy Auditor, which spawned this train of thought.

solind: @vanderpol the "all_exist" existence check was rendered essentially redundant with at_least_one_exists in an update of the OVAL specification when we abandoned the (poorly-documented) so-called "partial matching" scheme.  It might be possible to use sets and a set complement operation on the two variables to make it easier to see exactly which SIDs are missing configurations.  I'll have to think about it.

vanderpol: I don't think that reporting which SID's are missing configuration is going to be enough data for the average user to easily figure out what's going on.   If they just get a SID, we will get emails from every end user baffled.   They really need all of the metadata from the item.  We can make SCC report this data, but our goal being to promote this level of details from all OVAL/SCAP tools.  I should have started this conversation years ago.

solind: @vanderpol you could also use the filepath or username instead of the SID. Any unique field would do the job.

vanderpol: @solind thanks for that point.   I understand that this is an acceptable way to make existing tools report correctly and might be something worth investigating, but it also means rewriting and retesting all of the ntuser_tests in existence.  While not a huge number like Registry or File, there are a few hundred tests that would need to be redone.  Granted this might be faster than getting other OVAL/SCAP Scanners up to SCAP 1.4.

How about the thought of a behavior to "create_item_based_on_ntuser_file_existence"?   That would give control to the content author.  Default would be 'true' in order to not have to update all existing content though.

vanderpol: BTW this concept of a behavior would also be very useful for textfilecontent54 and xmlfilecontent, as you have to jump through similar convoluted methods with variables to check to ensure that all files that were in scope contain the required data, if the textfilecontent  pattern doesn't exist, the item isn't created.   Given the body of content that uses textfilecontent54 and xmlcontent, if we added behaviors to them, they would need to default to 'false' on creating items when the filepath exists.

My goal being making content easier to write and results be as intuitive as possible.  I fully agree that we 'can' do a lot of things with OVAL, but if it makes content harder to write, or results less understandable, then change/progress is needed.   OK off my soapbox.

solind: Adding a (non-default) behavior would probably be the least intrusive way to introduce such a change.  I'd personally call it something like "item_generation=(existing_values|every_ntuser)", with "existing_values" being the default.

The somewhat complicated approach I outlined above for writing the test is not unique to the ntuser_test.  It also applies to many of the various types of user tests that must both look for a user file, and also check its contents.  It's been so long that I can't recall every type of test where we had to use that approach, but there were at least several.

solind: Just a note -- choosing the default value is a matter of backwards-compatibility. It sounds like the existing content not implemented correctly to account for users lacking a specific user registry key.  To the extent that, for example, a Joval customer needed to have such an ntuser_check performed for compliance content, my bet is that it was implemented properly.  Adding a default behavior that creates partial matches might break that content.

Now, when I was in charge of this kind of thing and had a problem like yours, I'd also add a Joval property that customers could set to override the officially-specified behavior.  That's how you can have your cake and eat it too, which is the key to successfully writing enterprise software.

vanderpol: It sounds like we are on the same page. I'm in full agreement on the backward compatibility note.  I'll submit another round of updates tomorrow and see what you think.  Hoping we can get others to chime in as well.  Thanks again for your time and feedback.   I appreciate all of your efforts, keeping me honest.

oval-schemas/windows-definitions-schema.xsd
vanderpol: In reading the documentation for Win32_LoggedOnUser and Win32_LogonSession, it would appear to me that those are only to obtain data about the session that the application is running in, not to obtain information listing all currently logged in users.  

The Win32_LoggedOnUser association [WMI class](https://learn.microsoft.com/en-us/windows/desktop/WmiSdk/retrieving-a-class) relates a session and a user account.

The Win32_LogonSession WMI class (see [Retrieving a WMI class](https://learn.microsoft.com/en-us/windows/desktop/wmisdk/retrieving-a-class)) describes the logon session or sessions associated with a user logged on to a computer system running Windows.

If you have concerns about 'mandating' our method we can make it clear that it is just a recommendation not a requirement.  It's what we have been using for years, and just wanted to help new vendos not have to reinvent the wheel to find the data.

oval-schemas/windows-definitions-schema.xsd
vanderpol: In reading the Win32_NeworkLoginProfile documentation, I'm not confident that it contains the same data.  I'm updating documentation with clear information on how to get the data, and a reference to the Micrsoft documentation:  https://learn.microsoft.com/en-us/troubleshoot/windows-server/support-tools/scripts-to-retrieve-profile-age

oval-schemas/windows-definitions-schema.xsd
solind: > it would appear to me that those are only to obtain data about the session that the application is running in, not to obtain information listing all currently logged in users

I don't know exactly what you mean by this.  Application?  [Win32_LogonSession](https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-logonsession) definitely "describes the logon session or sessions associated with a user logged on to a computer system running Windows", and [Win32_LoggedOnUser](https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-loggedonuser) "relates a session and a user account".  It's the WMI way to determine who's currently logged in to a machine.  The keys loaded under HKU may tell you the same information, but I never found any documentation stating that conclusively.

oval-schemas/windows-definitions-schema.xsd
solind: It's hard to find documented means to determine some of this information, but [Win32_NetworkLoginProfile](https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-networkloginprofile) "represents the network login information of a specific user on a computer system running Windows".  It has datetime properties for LastLogon and LastLogoff.

Is the "profile age" really the same thing?  I don't know.  What happens if a user logs on when the machine is not connected to the domain?  Is the profile age updated?  Or is it updated only when group policy updates it on user logon?

oval-schemas/windows-definitions-schema.xsd
vanderpol: OK, I'll test it out internally and then update documentation to list either method.

oval-schemas/windows-definitions-schema.xsd
vanderpol: We are looking for the last time the logged into this specific computer and profile, which is what the LocalProfileLoadTimeHigh data provides.   Any time the profile is loaded it's updated.   The LastLogon and LastLogoff could be anytime that user logged into any computer in the domain?   I will have to test if the LocalProfileLoadTimeHigh is updated even when the system is disconnected from the network.   More discussions to follow.

oval-schemas/windows-definitions-schema.xsd
vanderpol: I logged into the same Windows Server via RDP with 2 remote users, and a another user was logged in at the console.  So there are 3 users concurrently logged into the system.

user1:  When running Get-WmiObject Win32_LoggedOnUser | Select-Object -Unique Antecedent 
- Results back are name="user1"
user2:  When running Get-WmiObject Win32_LoggedOnUser | Select-Object -Unique Antecedent 
- Results back are name="user2"

So it would appear from my testing that Win32_LoggedOnUser only returns data for the current user, not all users on the system.

When testing Get-WmiObject Win32_LogonSession, I see that there are 3 logins, but I do not see any straight forward way to determine the username or SID from any of the logged in sessions.  I get AuthenticationPackage, LogonID (which is not the SID), LogonType

In searching online, it seems that the quickest/easiest way to get the logged in users is to run 'quser', which returns back the username, but it doesn't seem entirely clear if they are local vs doman users, so doing a SID lookup might be tricky.

Back to HKEY_USERS, Microsoft documents it pretty well:
HKEY_USERS | Contains all the actively loaded user profiles on the computer. 
https://learn.microsoft.com/en-us/troubleshoot/windows-server/performance/windows-registry-advanced-users

I can add this documentation to the methods to make it clear it is a documented method, not just our hack.  Given that at least from our testing, reading the data from HKEY_USERS/\<SID\> is the only way to get to the data for logged on users, determining if the user is logged on from the same data seems straight forward.

oval-schemas/windows-definitions-schema.xsd
vanderpol: In my limited research, it appears that Win32_NetworkLoginProfile is tied to the last time you have logged into ActiveDirectory.  Here was my test:
1.  I was logged in from quite some time ago
2.  I disconnected my ethernet cable
3. I logged off and back into my computer
4. I ran Get-CimInstance -ClassName Win32_NetworkLoginProfile | Select-Object Name, LastLogon,
5. The data returns was 4/20/2026
6. I then reconnected my ethernet and ran the same command
7. Data was now correct with today datetime listed

At the same time, I exported off the HKLM\Software\Microsoft\Windows NT\CurrentVersion\ProfileList\ registry key as text, and exported again after logging in without eithernet, and saw that the data changed with each login/logoff.

And given that my current documentation matches the original pull request R&D, (which should have included this documentation) I feel pretty strongly that documenting the LocalProfileLoadTimeHigh method is 'correct'

https://github.com/OVAL-Community/OVAL/issues/221

oval-schemas/windows-definitions-schema.xsd
solind: It's actually not that simple.  See [this StackExchange question](https://stackoverflow.com/questions/23219718/powershell-script-to-see-currently-logged-in-users-domain-and-machine-status), particularly the answer(s) by [mjolinor](https://stackoverflow.com/users/574168/mjolinor).

solind: Perhaps mapping the subways of HKU works well enough ... that's just not how we determined who was logged in. We used that complex WMI method.

oval-schemas/windows-definitions-schema.xsd
solind: You're quite right, Win32_NetworkLoginProfile gets the last time the user logged into the domain. Sorry to lead you astray there!

oval-schemas/windows-definitions-schema.xsd
vanderpol: I added a note that other API's can be used, and mentioned Win32_LogonSession and Win32_LoggedOnUser

oval-schemas/windows-system-characteristics-schema.xsd
```diff
@@ -1692,6 +1711,7 @@
                               <xsd:element name="filepath" type="oval-sc:EntityItemStringType" minOccurs="0" maxOccurs="1">
                                    <xsd:annotation>
                                         <xsd:documentation>This element describes the filepath of the ntuser.dat file.</xsd:documentation>
+										<xsd:documentation>The existance of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
```
solind:
```suggestion
										<xsd:documentation>The existence of each ntuser.dat file determines the overall ntuser_item existence.</xsd:documentation>
```

solind (commented):
Spelling error, otherwise looks to include everything we discussed!

oval-schemas/windows-system-characteristics-schema.xsd
vanderpol: Good catch, and that line of documentation should not exist, as it's now defined by the new behavior, so shouldn't be there.  Removing and committing.

solind (approved):

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -3931,6 +3931,8 @@
                   <xsd:documentation>The ntuser test is used to check metadata associated with Windows ntuser.dat files. It extends the standard TestType as defined in the oval-definitions-schema and
                         one should refer to the TestType description for more information. The required object element references a ntuser_object and the optional state element specifies the ntuser
                         data to check.</xsd:documentation>
+						
+						<xsd:documentation>To ensure consistent results across OVAL interpreters, application developers should refer to documentation in the ntuser_item when implenting this test.</xsd:documentation>
```
A-Biggs, line 3935: `implenting` should likely be `implementing`

Issue #305, opened 2026-04-15
wmi57 does not support array type fields


balleman-ctr:
**Abstract**
Wmi57 uses the record datatype.  Field elements are then used for each field selected in the WQL query.  The field element only supports simple data types and therefore cannot represent complex data types such as arrays that may be returned by a WQL query.

**Proposal**
An interim step may be to update documentation on wmi57 to indicate this is an unsupported use and for tools to return an error when this is attempted.  A resolution may involve revisions to the record datatype to add support for complex data types.

**Additional context**
Current tool behavior is varied.  In testing with a field that is an integer array, SCC 5.14 returns text “ARRAY(address)” with a datatype of “int”, and Policy Auditor 6.5.9.144 crashes producing no results.

balleman-ctr:
Example Content
[oval_wmi_array_example.xml](https://github.com/user-attachments/files/26754719/oval_wmi_array_example.xml)

Test Results
[oval_wmi_array_example_scc514_results.xml](https://github.com/user-attachments/files/26754722/oval_wmi_array_example_scc514_results.xml)

vanderpol:
Thanks for reporting this issue, we are researching it on our side, and will make sure to bring it up in the next OVAL board meeting

vanderpol:
@maxullman, @A-Biggs or @solind any thoughts on this?  What does Joval/Arctic Wolf do with the above sample content?  Is this something we need to address in the OVAL schema, or just a bug in SCC and Trellix Policy Auditor?

vanderpol:
@balleman-ctr I'm not sure which DISA STIG you are trying to automate, but in looking at your query, I'm guessing you can re-use what we already have in our NIWC versions of the SCAP benchmarks, and are bundled with SCC 5.14.    I'll attached the standalone OVAL single that does this test using cmdlet vs WMI57.  This is by no means saying that your OVAL schema issue isn't valid, just trying to help your content automation.  Also, as I've mentioned before, you all are more than welcome to take ownership of any of our OVAL content that we have automated above and beyond what DISA has published.  No need to re-invent the wheel, and we can send it to you in any format the might be of use to you.

[Windows_10-11_credential_guard_must_be_running.xml](https://github.com/user-attachments/files/26788644/Windows_10-11_credential_guard_must_be_running.xml)

solind:
As I recall, for a complex WMIProperty, Joval would do something like convert it into a JSON-like string.

vanderpol:
[...]

PR #313, opened 2026-05-11
#305 - wmi57: clarify that only simple datatypes are supported

balleman-ctr:
This is proposed as in interim solution to #305 to specify that only simple datatypes are supported and that return of complex results should be reported as an error.


vanderpol (approved):
Looks good to me

solind (approved):
Looks good to me

Issue #311, opened 2026-05-11
Add behaviors to independent shellcommand to allow for status of 'error'


vanderpol:
In the current design of the independent shellcommand, any errors that occur when running arbitrary system commands are captured as part of standard out, but have no impact the status of the collected item.  This is by design, as certain commands, such as a find command may return back a bit of noise. 

Proposal:  Add a behavior to give the content author control of what standard errors mean, and if they should be displayed

For the behavior, some options are

- Include_stderror   (include standard error in the system data item, but it has no impact on the item's status, default)
- Ignore_stderror  (exclude all stderror messages, same as if command redirected stderror /dev/null)
- Item_error_if_stderror  (include standard error messages in the system data item, AND mark the system data item status as 'error')

We could also give the content author complete control and put in a stderror pattern in the object to allow content authors to filter and only see certain errors as they see fit.

Thoughts?

solind:
I recall feeling strongly that the item needed to have stderr lines that were distinct from the stdout lines, because they are completely different character streams and where one result will appear to be interleaved within another is completely the result of non-deterministic race conditions.  I'm guessing that's not what was decided, and now everyone can appreciate my wisdom LOL

solind:
This makes sense. In many ways, a Powershell script is just like any other script or executable -- the script's author has the ability to determine whether or not a runtime error results in a particular exit code. Therefore it makes sense to be able to control the desired behavior of the OVAL shellcommand, as the content-author and the script author may have different needs with regards to error control.

PR #314, opened 2026-05-13
Added behavior to shellcommand to allow for error status per #311

vanderpol:
Per discussions on #311 I've implemented the shellcommand behaviors


solind (approved):
Looks good to me.

PR #322, opened 2026-05-14
Added item_creation behavior for xmlfilecontent and textfilecontent54 per #315 and #316

vanderpol:
Added item_creation behavior, similar to new behavior in ntuser test which allows content authors to control when an item is created, and prevent false negatives when text/xml file may not contain the data required to populate an item, and would be overlooked.

### DIFF
--- a/oval-schemas/independent-definitions-schema.xsd
+++ b/oval-schemas/independent-definitions-schema.xsd
@@ -1916,6 +1916,19 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
                                     <xsd:documentation>'singleline' enables single line semantics in the regular expression provided by the pattern entity. This behavior is intended to align with the Perl regular expression 's' modifier: if true, the '.' metacharacter will match newlines. If false, it will not.  The default is false.</xsd:documentation>
                               </xsd:annotation>
                         </xsd:attribute>
+						<xsd:attribute name="item_creation" use="optional" default="all_object_elements_fullfilled">
+							  <xsd:annotation>
+									<xsd:documentation>For 'all_object_elements_fullfilled', items are only created when the entire object requirements are met (filepath, pattern, instance).</xsd:documentation>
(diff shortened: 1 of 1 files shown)


vanderpol:
FYI, @johnulmer-oval is implementing this in SCC, and we'll post sample results when he's done.

oval-schemas/independent-definitions-schema.xsd
```diff
@@ -2387,6 +2400,31 @@ SERVERPROPERTY('IsClustered') AS [is_clustered]
...
+            <xsd:complexContent>
+                  <xsd:extension base="ind-def:FileBehaviors">
+						<xsd:attribute name="item_creation" use="optional" default="all_object_elements_fullfilled">
+							  <xsd:annotation>
+									<xsd:documentation>For 'all_object_elements_fullfilled', items are only created when the entire object requirements are met (filepath, xpath).</xsd:documentation>
+									<xsd:documentation>For 'filepath_exists', items are created for each filepath on the system that matches the filepath/path/filname object requirement.  This option will prevent false negatives in instances where XML files exist that match the filepath object requirements, but do not contain the required xpath inside the file, which prevents item creation.</xsd:documentation>
+									<xsd:documentation></xsd:documentation>
+							  </xsd:annotation>
+							  <xsd:simpleType>
+									<xsd:restriction base="xsd:string">
```
g-burkholder, line 2419: It feels strange to have to define this enumeration twice in same schema.  But I don't see us defining named simpleTypes anywhere else, so I'm hesitant to start to do so now.  

Thoughts?  

```
<xsd:simpleType name="ItemCreationType">
    <xsd:restriction base="xsd:string">
        <xsd:enumeration value="all_object_elements_fullfilled"/>
        <xsd:enumeration value="filepath_exists"/>
    </xsd:restriction>
</xsd:simpleType>
```

vanderpol: I follow your concerns, although in looking at the windows definitions schema for example, and searching for something like "max_depth" or "recurse_direction" and you'll find several copies of the same thing.  I'm all for making it as simple and consistent as possible, but given our hard deadline to finalize OVAL 5.12.3 this week, I'm not sure now is the time to change course.

g-burkholder: No worries.  I think we're on the same page here.

g-burkholder (approved):

Issue #306, opened 2026-04-21
Update Windows ntuser test to collect data related to the user having logged in via Windows Explorer


vanderpol:
**Abstract**
There can be ntuser.dat files from accounts that cannot received GPO's and will never be able to comply with most security policies (service account profiles, profiles created from WinRM sessions etc... and the collected item for the ntuser test should include a new element with a name somthing like "has_logged_to_windows_explorer" which would a boolean that content authors could use to filter out undesirable ntuser.dat files, much like is currently done for enabled, and days since last login.

In each ntuser.dat from a human user that has logged in via Windows Explorer, will be a key called UserSignedIn located in:  Software\Microsoft\Windows\CurrentVersion\Explorer which will be a REG_DWORD of 1 or 0 to indicate if the user has logged in.

There might be other ways to determine this as well, but the SCC team has found this to be a reliable method for the past few years.

I would dare to say this is really a 'bug fix' more than a feature, so I'm interested in other OVAL board members feedback if this could be rolled into OVAL 5.12.3, and then be included in the upcoming SCAP 1.4.  If we include this in OVAL 5.13 as a feature, it will have to wait for SCAP 1.5

PR #323, opened 2026-05-14
Added new 'user_has_signed_into_explorer' element to ntuser state/item per #306

vanderpol:
Added new element to state/item which allows content authors to filter out users that have not logged in interactively and can be a source of false negatives, as this partial profile does not receive GPO's

### DIFF
--- a/oval-schemas/windows-definitions-schema.xsd
+++ b/oval-schemas/windows-definitions-schema.xsd
@@ -4086,7 +4086,7 @@
 								  </xsd:element>
 								    <xsd:element name="user_has_signed_into_explorer" type="oval-def:EntityStateBoolType" minOccurs="0">
                                           <xsd:annotation>
-                                                <xsd:documentation>The user_has_signed_into_explorer element describes if the user account has ever interactively logged into Windows Explorer.  This is important as User GPO's are applied when a user logs into Windows Explorer, and are not applied for non-interactive logins such as SSH or WinRM.  Those non-interactive logins may cause false positives, and content authors may want to filter them out.</xsd:documentation>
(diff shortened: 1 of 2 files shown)


solind (commented):
Can't this test already just be implemented as a relatively straightforward ntuser_test?

oval-schemas/windows-definitions-schema.xsd
```diff
@@ -4084,6 +4084,12 @@
 											<xsd:documentation>For more information, refer to https://learn.microsoft.com/en-us/troubleshoot/windows-server/support-tools/scripts-to-retrieve-profile-age</xsd:documentation>
 									   </xsd:annotation>
 								  </xsd:element>
+								    <xsd:element name="user_has_signed_into_explorer" type="oval-def:EntityStateBoolType" minOccurs="0">
+                                          <xsd:annotation>
+                                                <xsd:documentation>The user_has_signed_into_explorer element describes if the user account has ever interactively logged into Windows Explorer.  This is important as User GPO's are applied when a user logs into Windows Explorer, and are not applied for non-interactive logins such as SSH or WinRM.  Those non-interactive logins may cause false positives, and content authors may want to filter them out.</xsd:documentation>
```
balleman-ctr: I’m getting hung up on a few specific parts of the documentation (first paragraph).  I’ve attempted a new version for discussion:

The user_has_signed_into_explorer element describes if the user account has ever run Windows Explorer. This is a practical indicator of accounts using interactive (desktop/GUI) sessions and is not set by non-interactive logon methods such as WinRM or SSH. Content authors may use this element to exclude non-interactive users from user policy checks.

vanderpol: @balleman-ctr thanks for your feedback, and I like your revised verbiage, I'll update the schema and ask for a re-review.

A-Biggs: I'm not positive, but I think WinRM logons may get GPOs applied, as a while back the team was doing work around PowerShell and GPOs to add restrictions and I believe the GPOs were having an impact.  The same user may have also logged in using non-WinRM methods, and that is how the GPOs were applied to the profile, but I can't recall, nor do I think we paid attention to that at that time.

A-Biggs (approved):

PR #325, opened 2026-05-21
Add support for ipv6 to unix interface test to resolve #312

vanderpol:
Committing @johnulmer-oval update to resolve #312

Attached are sample results
[Linux_OVAL-Results_unix-def_interface_test.xml](https://github.com/user-attachments/files/28105264/Linux_OVAL-Results_unix-def_interface_test.xml)


solind (commented):
I don't think we've discussed this previously, but this is a pretty significant change in the behavior of this object. Given that, the better course of action here would be to deprecate the interface_test and introduce a new interface512_test that had distinct collection entities for ipv4 and ipv6 addresses.

johnulmer-oval (changes requested):
Suggest moving this to 5.13 so we can more deliberately handle the multiple IPs of both v4 and v6 flavors appropriately.

PR #328, opened 2026-05-22
Fixed typo in draft 5.12.3 schemas on the deprecation of up

vanderpol:
Just need to fix the close tag on the up enumeration to be after the deprecation annotation.


johnulmer-oval (approved):
Looks good.  Annotation chunks now fall inside the 'enumeration' XML chunks to which thay belong.

PR #330, opened 2026-05-26
Deprecated the 'windows_view' element and attribute per #329

vanderpol:
Documented the unused 'windows_view' as deprecated in 5.12.3 and to be removed in 6.0.1


balleman-ctr (approved):
I don't see any problems with merging.

Two observations that could be addressed separately and more generally:
- these deprecations do not add schematron, but some others have it
- there are some indentation inconsistencies

g-burkholder (approved):
We aren't using windows_view in our active content, and I'm not aware of any intent to use it in the future.  I concur with deprecating it.

PR #335, opened 2026-06-05
Merge 5.12.3 develop branch back into master for official 5.12.3 release

vanderpol:
All PR's have been approved, just merging the 5.12.3 develop to master.


guidelines/oval-schema-documentation/independent-definitions-schema.rst
```diff
@@ -780,6 +783,28 @@ Child Elements
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. list-table:: Attributes  
+    :header-rows: 1  
+  
+    * - Attribute  
+      - Type  
+      - Desc.  
+    * - error_if_exit_status_not_0  
+      - xsd:boolean (optional *default*='false')  
+      - 'error_if_exit_status_not_0' enables the OVAL interpeter to set the shellcommand item to a status of 'error' if exit_status is not set to 0. The default is false.
```
solind: Technically, what we want is to set the oval_results object status to error.

guidelines/oval-schema-documentation/independent-definitions-schema.rst
```diff
@@ -780,6 +783,28 @@ Child Elements
...
+  
+    * - Attribute  
+      - Type  
+      - Desc.  
+    * - error_if_exit_status_not_0  
+      - xsd:boolean (optional *default*='false')  
+      - 'error_if_exit_status_not_0' enables the OVAL interpeter to set the shellcommand item to a status of 'error' if exit_status is not set to 0. The default is false.  
+    * - error_if_stderr_exists  
+      - xsd:boolean (optional *default*='false')  
+      - 'error_if_stderr_exists' enables the OVAL interpeter to set the shellcommand item to a status of 'error' if data is returned to stderr. The default is false.
```
solind: _ibid_

guidelines/oval-schema-documentation/independent-definitions-schema.rst
```diff
@@ -1223,6 +1248,11 @@ Attributes
     * - singleline  
       - xsd:boolean (optional *default*='false')  
       - 'singleline' enables single line semantics in the regular expression provided by the pattern entity. This behavior is intended to align with the Perl regular expression 's' modifier: if true, the '.' metacharacter will match newlines. If false, it will not. The default is false.  
+    * - item_creation  
+      - Restriction of xsd:string (optional *default*='all_object_elements_fullfilled') ('all_object_elements_fullfilled', 'filepath_exists')  
+      - For 'all_object_elements_fullfilled', items are only created when the entire object requirements are met (filepath, pattern, instance).
```
solind: Should also note that we want the oval_results object status to be "exists" vs "does not exist", which dictates whether or not the associated items are ever actually compared to a state.

solind (changes requested):
This isn't necessarily an exhaustive review, but I did notice three documentation issues.

guidelines/oval-schema-documentation/independent-definitions-schema.rst
johnulmer-oval: @Solind, is it not the case, that an item of status 'error' will necessarily result in an object status of 'error'?    Looking  at it more broadly, the 'shellcommand' test has been a little out of ordinary as it always produces an item with status of 'exists'.  The 'error_if_exit_status_not_0' behavior, when 'true', will make the collection of a shellcommand object behave (hate to use that overloaded word, here) more like a 'typical' OVAL object.

guidelines/oval-schema-documentation/independent-definitions-schema.rst
johnulmer-oval: Kinda' similar to my comment, above, doesn't OVAL already require that item existence determines object status?  I think if an item 'exists', the associated object necessarily must reflect that?  No?

solind: There is no association between the definition object and the system-characteristics item but for system-characteristics object (which is what I erroneously called the result object in my last comment). It is the bridge that links the two together.

If you look at the OVAL evaluation rules, it's the sc-object status that determines the existence check result. There is no comparison between states and items unless that existence check passes.

What I didn't like about the documentation is that it only refers to items.  But these new behaviors are essential to determining the sc-object status, in addition to whether or not sc-items are created for that sc-object. It's a nuance, but it's the kind of nuance that leads to confusion down the line, when this conversation is but a distant memory.

solind: Incidentally I either found the .rst diffs much more readable, or those changes appeared earlier than the corresponding XSD changes, which is why I put the comments where I did.

vanderpol: Thanks @solind I generated the .rst files today, after we pushed the .xsd PR into 5.12.3 develop branch,  If @johnulmer-oval  ends up making changes to the .xsd files, I'll regenerate the .rst files.

johnulmer-oval: Would it suffice to add the language?

"If the 'item_creation'  behavior causes collection of an OVAL item with status of 'exists', the associated system_characteristics object(s) will reflect the existence of that collected item."

I'm likely not being adequately pedantic. Give me some editing it that does communicate your point adequately.

guidelines/oval-schema-documentation/independent-definitions-schema.rst
solind: OK, actually what we really want is the sc:object flag attribute to be set to "error".

Again, in the sequence of the evaluation, I believe what you want is for the def:test existence_enumeration check to result in an error in the case of a non-zero exit code.  Otherwise, if there is no state in the test, then no error will be generated.

guidelines/oval-schema-documentation/independent-definitions-schema.rst
solind: It's more important to note specifically that the sc:object will have a flag of "does not exist" if there are no resulting items with collected entities for filepath, pattern AND instance.  That is, if the file exists but not the pattern, with the all_object_elements_fullfilled behavior, the object would be said to not exist.

solind (commented):
Sorry for the tardy replies, I had to help move my son out of his dorm!

vanderpol:
@solind we have to get OVAL 5.12.3 finalized before Monday AM, as NIST is publishing SCAP 1.4, and the URL has to be live with the 'official' version by then.   If you feel your recommended documentation updates warrant an update, please commit your updates, and I'll roll them into the final release.  Given that these are just documentation changes, if we don't get it updated today, I'm going to proceed with releasing OVAL 5.12.3 and we can update documentaiton for 5.12.4 or 5.13.   We can also ensure that the intent of your documentation requirements are tested via the SCAP 1.4 self assertion content.

